PR #22, opened 2026-07-28
Fall back to official arXiv RSS

ktwu01:
## Root cause

The controlled post-merge Daily Radar run failed before persistence because arXiv was required and returned no records. Direct inspection showed the Atom endpoint returning HTTP 429 or timing out from shared hosted-runner IPs, while the official arXiv category RSS endpoints were healthy.

Failed run: https://github.com/ktwu01/benchmark-radar/actions/runs/30323063835

## Fix

- Use official `rss.arxiv.org` category feeds as the production path on GitHub-hosted runners and filter relevant benchmark/evaluation/data records locally.
- Retain the Atom parser—including `updated` handling—for deployments with stable egress that explicitly enable it.
- Preserve stable arXiv IDs, announcement timestamps, update/release event kinds, authors, overlap filtering, and the per-source cap.
- Keep `required: true` strict. Required-source failures now identify whether a source failed or returned zero records.
- Add coverage for both HTTP-429 fallback and RSS-primary operation.

## Validation

- `ruff check .`
- `ruff format --check .`
- `pytest -q` — 41 passed
- Live official RSS collection — 60 relevant arXiv records
- Isolated full collector run — 30 ranked evidence, 18 attention, arXiv healthy with 60 raw records, two-day dashboard generated
- `git diff --check`

OpenAlex and Brave remain separately blocked by absent repository API-key secrets; this change does not mask those optional-source health failures.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -41,6 +41,21 @@ sources:
     # timestamp is submission time, while announcements pause on weekends.
     overlap_hours: 120
     request_delay_seconds: 3
+    # GitHub-hosted runners are routinely rate-limited by the Atom API's
+    # shared-IP quota. Use official category RSS feeds and filter locally.
+    # Atom parsing remains available for deployments with a stable egress IP.
+    atom_enabled: false
+    rss_categories: [cs.AI, cs.CL, cs.CV]
+    rss_keywords:
+      - benchmark
+      - evaluation
+      - evaluator
+      - leaderboard
+      - dataset
+      - test set
+      - data contamination
+      - benchmark leakage
+      - data quality
     queries:
       - 'all:"large language model" AND (all:benchmark OR all:evaluation)'
       - '(all:benchmark OR all:dataset) AND (cat:cs.AI OR cat:cs.CL OR cat:cs.CV)'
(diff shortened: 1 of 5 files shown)


horatiol (commented):
Screened `sources.py` and `pipeline.py` for performance-sensitive surface.

The new `_fetch_arxiv_rss()` path makes N sequential HTTP GETs (N = `len(rss_categories)`, default 3: cs.AI, cs.CL, cs.CV). The previous Atom path made `len(queries)` sequential requests with `time.sleep(request_delay_seconds)` between them. With `atom_enabled: false` (the production config), the new code skips Atom entirely and goes straight to RSS with no inter-request sleep. That is a net improvement in wall time for the production path.

The local keyword filtering — `any(keyword in f"{title} {description}".casefold() for keyword in keywords)` — allocates one string per RSS item but over a feed of at most a few hundred entries this is negligible. The `dict`-keyed deduplication across categories is O(1) per item. The final sort and `[:limit]` slice are fine.

The `pipeline.py` change replaces a sorted comprehension with an explicit loop for error message construction. This runs once per source health check at pipeline teardown, not on any hot path.

This is a scheduled batch pipeline with no user-facing latency budget. No measured regression and no structural performance concern in this diff.

— Performance Engineer, Liftoff

Issue #53, opened 2026-07-30
Expose stale-data state when today's snapshot is missing


ktwu01:
## Observation

The repository currently has no `data/snapshots/2026-07-30.json` and no daily issue for 2026-07-30. The newest canonical snapshot is `data/snapshots/2026-07-29.json`, and the newest daily report is #47.

Sources:
- https://koutian.is-a.dev/benchmark-radar
- https://github.com/ktwu01/benchmark-radar/blob/main/data/snapshots/2026-07-29.json
- #47

## Why this matters

The dashboard should distinguish:

1. a successful run with zero new records;
2. a run that has not occurred;
3. a collection or publication failure.

Without this distinction, an automated日报 may relabel yesterday's records as today's additions.

## Proposed changes

- Publish `last_successful_collection_at` and `last_successful_publication_at`.
- Show a stale-data banner when the selected date lacks a canonical snapshot.
- Publish a zero-change daily issue when a run succeeds but finds no additions.
- Surface connector and publication failures separately.
- Require each daily issue to reference a same-date snapshot or state that none exists.

## Acceptance criteria

- [ ] Dashboard shows latest snapshot date and age.
- [ ] Missing-run, failed-run, and zero-change states differ.
- [ ] Daily reports cannot silently reuse the previous day's snapshot.
- [ ] Automated consumers can determine freshness from structured fields.

Issue #52, opened 2026-07-30
AI Benchmark Landscape Report: Market Scope, Agentic Evaluation, and Coverage Gaps


ktwu01:
## Objective

Turn issue #52 into a versioned AI benchmark landscape report for evaluation researchers, benchmark maintainers, and research leaders. The report will separate the installed benchmark stock from the live flow of new and updated artifacts.

The current analysis is in [`docs/reports/ai-benchmark-landscape-report.md`](https://github.com/ktwu01/benchmark-radar/blob/main/docs/reports/ai-benchmark-landscape-report.md).

## What the report will contain

1. Reader brief with findings and confidence labels.
2. Registry-based market census with definitions, overlap, and uncertainty.
3. Live activity from the daily radar, reported separately from the census.
4. Landscape map by domain, modality, evaluation target, property, maturity, and access.
5. Agentic evaluation deep dive covering environments, tools, memory, safety, and task domains.
6. Adoption view for runnable artifacts, downloads, stars, citations, and hosted leaderboards, kept source-specific.
7. Measurement audit covering connector health, source caps, synthetic dates, classifier performance, and identity gaps.
8. Research agenda and reproducibility appendix.

## Findings from the current corpus

- 645 distinct artifacts from 791 sightings.
- Current overlapping tags: 512 benchmark, 450 dataset, 401 evaluation, 78 agentic, and 15 data quality.
- The agentic count was repaired from 3 to 16 by historical re-scoring, then to 78 by the measured proximity classifier.
- The corpus is a discovery-flow instrument, not a market census: four of nine dates are simulated, only three source families contribute artifacts, and 19 of 20 canonical benchmarks tested are absent.
- Agentic evaluation is concentrated in software and computer use, tool use and planning, memory and long-horizon behavior, professional tasks, and security.

## Acceptance criteria

- Define benchmark, artifact, family, stock, and flow.
- Keep census and live-flow figures in separate tables and visualizations.
- Enumerate more than one registry ecosystem.
- Human-label a random sample of at least 100 benchmark-tagged artifacts for precision, with an interval.
- Obtain domain-expert review of the agentic labels and validate them on a later window.
- Link papers, repositories, datasets, leaderboards, and result stores when exact evidence supports a shared family.
- Attach a source, cutoff date, and confidence label to every headline number.
- Exclude simulated and connector-incomplete dates from trend estimates.
- Rebuild the report from committed snapshots and versioned analysis code.

## Implementation sequence

[...]

Issue #42, opened 2026-07-28
map improvement


ktwugoat:
https://koutian.is-a.dev/benchmark-radar/?view=map

This map, we have more than 200 things, but in the artifacts, we only have a dozen. And as for the organizations, we only have this. 

So this is just ridiculous.

1140 cumulative entities · 216 observations · 1560 relationships · showing 16 ranked artifacts

I just didn't see the aim of doing this map because it doesn't seem to reflect any real insight. It's not insightful at all, I think.

Issue #35, opened 2026-07-28
Simulate historical snapshots to reach 30-replay coverage


ktwu01:
## Context

Follow-up from #8. The 30-snapshot deterministic replay acceptance criterion was checked off in #8's body while only 2 real snapshots exist (`data/snapshots/2026-07-27.json`, `2026-07-28.json`). Waiting ~30 days of daily runs to satisfy this is unnecessary: connectors (arXiv, GitHub, Hugging Face) can be queried against historical windows, so we already have enough source data to construct the missing days without waiting on the calendar.

> 但过往几天的你有全部的data也可以做趋势吧，甚至整个history你都有充分的数据可以做

## Proposal

- Add an `--as-of DATE` (or equivalent) mode to the collector that:
  - Runs each connector constrained to return only records visible as of that simulated date (no data dated after it leaks in).
  - Writes a snapshot file for that simulated date under `data/snapshots/` in the existing flat `YYYY-MM-DD.json` layout.
- Backfill enough simulated daily snapshots (working backward from today or forward from the earliest available source data) to reach 30 total snapshot files.
- Run `benchmark-radar backfill` / `rebuild` (`src/benchmark_radar/cli.py`) against the resulting 30 snapshots and confirm aggregates replay deterministically.

## Open question to resolve first

Confirm which connectors actually support a historical/as-of query (arXiv has date-range search; GitHub search API supports `created:`/`pushed:` filters; Hugging Face Hub API support needs checking). If a connector can't be constrained to a historical cutoff, note it explicitly as a known gap rather than silently including leaked future data.

## Acceptance criteria

- [ ] Collector supports simulating a historical "as-of" run per connector, with no future-relative-to-simulated-date data included
- [ ] 30 daily snapshot files exist under `data/snapshots/`
- [ ] `benchmark-radar backfill` replays all 30 snapshots to identical, deterministic aggregates
- [ ] Any connector that cannot support historical simulation is documented as a known limitation

ktwugoat:
https://koutian.is-a.dev/benchmark-radar/?view=trends , we can filter for those that have a release. That way, we can filter by when they released, and after that, we will get this.

And as for those tag= updated, that should not be considered in the 30 days change.

ktwu01:
not so useful

PR #56, opened 2026-07-30
Fix issues #53, #52, #35: freshness banner, corpus totals, history backfill

ktwu01:
## Summary

- **#53** Dashboard freshness fields (`last_successful_collection_at`, `degraded`) derived from required-source health only (arxiv/huggingface/github), plus a staleness banner for when the latest snapshot is stale (>30h) or a required connector failed. Optional sources like brave/openalex, which fail on every run without an API key, are deliberately excluded from the "degraded" signal so it doesn't fire constantly.
- **#52** A "Corpus totals" panel on the Today view surfaces `corpus.aggregates.topics` (already computed, but previously only reachable via the trend map that #42 flags as unreadable) as total distinct artifacts by category (benchmark/dataset/evaluation/data quality).
- **#35** A `simulate-history` CLI command backfills historical snapshots by fetching each eligible connector once (not once per simulated date) and re-scoring the pool per simulated day. arXiv and Brave are excluded and documented as known limitations (arXiv's reliable RSS path has no date-range query; its Atom path is already disabled for rate-limit fragility; Brave's index reflects live crawl state, not historical availability). Ran it against live GitHub/HF APIs: only 4 of 26 candidate days had reachable records (both APIs are recency-sorted with no per-day cursor), so this reaches 8 total snapshots rather than the originally scoped 30 — the other 22 days are explicitly skipped and reported, not silently published as empty.

## Test plan

- [x] `pytest` — 151 passed
- [x] `ruff check` / `ruff format --check` — clean
- [x] `benchmark-radar backfill` run twice, `site/data/radar.json` byte-identical (deterministic replay)
- [x] Verified `degraded`/`required_coverage_gaps` correctly reflect the arXiv gap on simulated days without falsely flagging the latest (real) day as degraded
- [ ] Could not visually verify the dashboard banner/panel in a real browser (headless Chrome unavailable in this sandbox) — static review of HTML/CSS/JS only; would appreciate a manual look before merge

https://claude.ai/code/session_01NnR4FqkJRiafnP7Uj4CbJx

### DIFF
--- a/data/snapshots/2026-07-23.json
+++ b/data/snapshots/2026-07-23.json
@@ -1,0 +1,909 @@
+{
+  "attention": {
+    "observations": []
+  },
+  "date": "2026-07-23",
+  "discovery_state": {
+    "attention": {
+      "benchmark-social-signal:hacker-news:48989306": {
+        "last_seen_at": "2026-07-27T17:58:20.350823+00:00",
+        "observed_at": "2026-07-27T17:58:20.350823+00:00"
+      },
+      "benchmark-social-signal:hacker-news:48991245": {
+        "last_seen_at": "2026-07-27T17:58:20.350823+00:00",
+        "observed_at": "2026-07-27T17:58:20.350823+00:00"
+      },
+      "benchmark-social-signal:hacker-news:48994388": {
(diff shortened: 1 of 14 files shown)


ktwugoat:
good

PR #77, opened 2026-08-01
Collect OpenAlex without an API key

ktwu01:
## Summary

- stop requiring `OPENALEX_API_KEY`: OpenAlex is a fully open API with no free key to obtain
- identify the caller with the `mailto` contact address OpenAlex uses to route requests into its faster "polite pool"
- still forward a key when one is set, since OpenAlex sells premium keys
- strip the key so a blank or newline-padded secret is dropped rather than sent

## Why

`fetch_openalex` opened with:

```python
api_key = os.getenv("OPENALEX_API_KEY")
if not api_key:
    raise RuntimeError("OPENALEX_API_KEY is not configured")
```

No free OpenAlex key exists, so this guard could never be satisfied and the connector disabled itself on every scheduled run. The verified production snapshot `data/snapshots/2026-08-01.json` records `openalex` as `ok=false, item_count=0` with exactly that error. The run only stayed green because `openalex` is not marked `required`.

## Validation

Checked against the live API rather than assumed:

| Request | Result |
|---|---|
| no key, no mailto | HTTP 200 |
| no key + `mailto` | HTTP 200, real results |
| `api_key=bogus` | HTTP 401 |
| `api_key=<space>` | HTTP 401 |
| `api_key=bogus\n` | HTTP 401 |

That 401 behavior is why the value is stripped: a wrong or blank key is rejected, not ignored, so forwarding an unset secret would have re-broken the source through a different path.

- live keyless fetch using the committed config returns **68 works**
- same fetch with `OPENALEX_API_KEY="   "` also returns **68 works**
- `pytest` — 182 passed
- `ruff check` / `ruff format --check` clean

## Review

Codex flagged the unstripped key as a P1 (unset Actions secrets expand to `""`, pasted keys carry newlines). Confirmed against the live API and fixed in 52d7713, with parametrized regression coverage for `""`, `"   "`, `"\n"`, `" \t\n"` and a trailing-newline real key.

Gemini review was attempted first per repo policy, but the `gemini` CLI is unauthenticated in this environment (no `oauth_creds.json`), so Codex was used as the reviewer instead.

## Note

This restores a source that returns future-dated works (for example `publication_date: 2050-01-01`). Recency scoring clamps age at `max(0.0, ...)`, so a future date scores a permanent 100. That is pre-existing and independent of this change; filed separately rather than widened into this PR.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -235,6 +235,10 @@ sources:
 
   openalex:
     enabled: true
+    # OpenAlex has no free API key. It asks callers to identify themselves with
+    # a contact address, which routes requests to the faster and more reliable
+    # "polite pool". This is a public repository address, not a credential.
+    mailto: ktwu01@users.noreply.github.com
     searches:
       - large language model benchmark
       - artificial intelligence evaluation dataset
--- a/src/benchmark_radar/sources.py
+++ b/src/benchmark_radar/sources.py
@@ -645,15 +645,24 @@ def fetch_github_releases(
 
 
 def fetch_openalex(config: dict[str, Any], since: datetime, limit: int) -> list[RadarItem]:
-    api_key = os.getenv("OPENALEX_API_KEY")
-    if not api_key:
-        raise RuntimeError("OPENALEX_API_KEY is not configured")
+    # OpenAlex is a fully open API: anonymous requests succeed, and there is no
+    # free key to obtain. Requiring one disabled a working source on every run.
+    # A key is still forwarded when set, because OpenAlex sells premium keys,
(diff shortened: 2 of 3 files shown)


ktwu01:
Superseded by #87. OpenAlex changed its documented production access model in February 2026: free API keys replaced the former `mailto` polite-pool path. The replacement PR requires and normalizes the free key, retains the date-bound fix, quarantines future records without sacrificing source caps, and includes persistent source-health warnings. Closing this PR rather than merging an implementation built on the outdated keyless premise.

PR #92, opened 2026-08-02
Read the Fable 5 / Mythos 5 benchmark table and order the registry by date

### LINKED ISSUE #90: Update benchmark-radar ?view=leaderboard: Update Claude Fable/Mythos 5 benchmarks and sort registry model cards by date
improve benchmark-radar/?view=leaderboard
Claude Fable 5 and Mythos 5 这个release 里面：你漏掉了一些benchmark。你需要OCR这个图retrieve所有mentioned benchmarks in 这个release。然后，Model cards in the registry 这个要按照时间排序。

https://www.anthropic.com/_next/image?url=https%3A%2F%2Fwww-cdn.anthropic.com%2Fimages%2F4zrzovbb%2Fwebsite%2F1e65982497d7d4891219ed0e83141625a291b860-2600x2870.png&w=3840&q=75

ktwu01:
Fixes #90.

## The image table

The release states its headline results in an image, so the pass that built this card could only see the benchmarks named in prose. It recorded five, two of which appear nowhere in the release.

Reading the table directly gives 13 benchmarks. Combined with the three named only in prose, the card goes from 5 to 16.

| | Benchmarks |
|---|---|
| Added from the table | SWE-bench Pro, FrontierCode, GDPval, GDP.pdf, Blueprint-Bench 2, AutomationBench, OSWorld, Legal Agent Benchmark, HLE, BioMysteryBench, Terminal-Bench, ExploitBench, HealthBench Professional |
| Kept, named in prose | Frontier-Bench, CursorBench, ViBench |
| Removed, absent from the release | BrowseComp, SWE-bench Verified |

Two things the transcription turned up:

- **FrontierCode and Frontier-Bench are two different Cognition evaluations.** The release scores them separately. Merging them on the shared prefix would double one adoption and erase the other.
- **ExploitBench carried a wrong release date.** It was dated from the GPT-5.6 preview it links to, which post-dates a table that measures against it. That date is now unset rather than re-guessed, along with seven others (see below).

Five of the new benchmarks are vendor-internal, with no published task specification. They carry no URL, and their caveats say why the score cannot be reproduced rather than leaving the gap unexplained.

## Ordering

`model_cards` sorted by publisher then date, which grouped each vendor's whole history together and left the newest document in the registry halfway down the list. It now sorts newest first.

Details worth noting:

- Descending needs a `_descending` wrapper, not `reverse=True` — the latter would also flip the tie-breakers to Z-to-A.
- Undated cards sort last. An unknown date is not evidence of recency.
- The key ends in `document_type` and the id. Z.ai's GLM-5 model card and technical report tie on every other key, so without them the roster followed YAML file order.
- The dashboard had two more copies of the old ordering (the "Registry coverage" list, and the per-organization roster). All three now agree, and the JS comparator compares by codepoint so it cannot diverge from the Python sort on case.

## Test plan

- `pytest -q` — 255 pass (6 new tests)
- `ruff check .` and `ruff format --check .` — clean
- Each new test verified to fail against the pre-change behavior, so they are real regression guards
- Output is byte-identical across `PYTHONHASHSEED` 0/7/99
- The JS comparator is extracted verbatim from `app.js` and checked against the generated `radar.json`: both rosters produce identical order

## Not done

**No browser visual review.** This environment has no display, and headless Chrome could not render the 7.9MB `radar.json`. The JS was verified under Node against real data, but nobody has looked at the page. Worth a visual check on the leaderboard view before merge.

### DIFF
--- a/data/model_cards.yml
+++ b/data/model_cards.yml
@@ -722,14 +722,115 @@ benchmarks:
 
   - id: exploitbench
     name: ExploitBench
-    aliases: ["ExploitBench"]
+    aliases: ["ExploitBench", "ExploitBench (Cap%)"]
     domain: security
     url: https://deploymentsafety.openai.com/gpt-5-6-preview
-    released: 2026-06-26
+    # Dated to the earliest document observed reporting it, Anthropic's
+    # 2026-06-09 Fable 5 / Mythos 5 comparison table, not to the GPT-5.6 preview
+    # the URL points at. The previous 2026-06-26 date was that preview's own
+    # publication date, which is a claim about when OpenAI reported the
+    # benchmark rather than when the instrument existed, and it post-dated a
+    # card that demonstrably measures against it.
+    released: 2026-06-09
     caveat: >-
       Offensive-security capability measured as a risk indicator against a
-      preparedness threshold, not a leaderboard to top. Only OpenAI has
-      published against it so far.
+      preparedness threshold, not a leaderboard to top. Reported as a
+      capability percentage, and no public specification of the task set has
+      been published, so the number cannot be independently reproduced.
+
+  # ---------------------------------------------------------------------------
+  # Instruments first observed in Anthropic's 2026-06-09 Claude Fable 5 /
+  # Mythos 5 comparison table (issue #90). That table is published as an image,
+  # so these names are absent from the page text and were missed by every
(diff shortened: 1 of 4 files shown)


ktwu01 (commented):
ok

PR #164, opened 2026-08-08
Survey who builds the popular benchmarks

### LINKED ISSUE #156: @Slim shilin popular bench authors profile info
@Slimshilin said

能不能跟进一步做个平行的事情，吧啦一下所有popular bench（怎么popular你来定），尤其是开源大team的authors，他们的对应个人profile信息，看看有没有在做数据业务的

---

## Redirect, after audit (2026-08-19)

Two deliveries, two rejections. Both rejections were about the same thing, and it was not crawl coverage.

### What went wrong

The 2026-08-14 delivery ranked benchmark repos by **GitHub stars**. That is why the 2026-08-18 comment reads:

> 这100个看着不行啊，tb和swebench都没统计在最上面嘛？？？
> princeton-nlp/SWE-bench 其实是在 repos_named.csv 里面的，但是 somehow the snapshot in previous comment里面没有列出来。
> terminal bench 也确实没在table里面

SWE-bench was in the file and did not surface. terminal-bench was absent entirely. Neither is a coverage bug. Stars measure how many people bookmarked a repo, which is not the same as how important the benchmark is, and the famous LLM-eval benchmarks are exactly the ones this diverges on. Re-running the crawl with a wider net will not fix it, because the ordering is what is broken.

### The correct ranking signal already exists

**Rank by how prominently the benchmark appears on the leaderboard page**, i.e. by how many model publishers actually report it. That is already computed and already rendered at https://koutian.is-a.dev/benchmark-radar/?view=leaderboard — the comment says so directly:

> we have lots of FAMOUS benchmarks used by lots of LLM vendors BUT didn show above in comments.

So the input is not a new crawl. It is the ranking the site already publishes.

### The deliverable

A list of **people**, ordered by the fame of the benchmark they work on, most famous first. Not a list of repos, not a contributor count, not "top 40".

- [ ] Take the benchmark ordering from the leaderboard page (vendor-reported prominence), not from stars
- [ ] SWE-bench and terminal-bench appear at or near the top, or the ranking is still wrong
- [ ] For each benchmark, name its authors/maintainers with a profile link, so a reader can see the person at a glance
- [ ] Keep the data-business flag from the earlier pass, but as a column, not as the sort key

On the original data-vendor question: the 2026-08-14 audit found that of 11 leads, essentially one (Lintang Sutawika, Datasaur/KonvergenAI) actually runs a data-annotation business, and the famous LLM-eval benchmarks are academic and model-lab work rather than data vendors. That is a real finding and it stands. It is also why sorting by "is this person a data vendor" produces a list of one, and sorting by benchmark fame produces a useful list.

### On 哪些信息是应该放在最显眼地方的

The closing note was not a request for more rows. It was a request that the important thing be visible first. Same complaint as #227 and #280, applied to a table instead of a page.

### Done when

- [ ] The list is ordered by benchmark prominence taken from the leaderboard page, not by GitHub stars
- [ ] SWE-bench and terminal-bench appear at or near the top; if they do not, the ranking is still wrong and this does not close
- [ ] Each row names a person with a profile link, so the reader sees who rather than how many
- [ ] The data-business flag is present as a column, not as the sort key

Two deliveries have already been rejected on the ordering. A third delivery that changes the row count but not the ordering does not close this.

ktwu01:
Closes #156.

## The seed list already existed

The corpus tracks **6,060 people and 741 organizations** through `AUTHORED_BY` edges, built by `build_corpus` and never used for anything. This puts it to work.

`benchmark-radar authors` ranks the GitHub benchmark repositories the radar has seen, reads their public contributor lists, and records the profile fields each contributor chose to publish (name, company, blog, bio, location, follower count).

## "Popular" is derived, not asserted

The issue leaves the definition to us, so it is computed from the corpus and is reproducible: **stars, then days tracked by the radar, then connector breadth**, capped at three repositories per owner so one prolific org cannot crowd out every other team.

Repository selection needed the same guard the briefing already has. A lone `evaluation` tag admitted `santifer/career-ops` (63k stars, a careers repository) as a benchmark, so a repository now needs an explicit `benchmark`/`dataset` category or a second corroborating tag. After the fix the seed list is `huggingface/datasets`, `NVIDIA/cosmos`, `yzhao062/pyod`, `torchgeo/torchgeo`, `modelscope/evalscope`, and similar.

## Data work is read from the author's own words

The classifier matches the self-written `bio` and `company` text against data-work signals and **never infers from a repository name**, so an author is credited because they described the work themselves. On a live run against the real API it found:

- `lhoestq` — "ML Engineer @huggingface | Maintainer of 🤗Datasets"
- `kritinv` — "Co-founder @ Confident AI (YC W25) | LLM Evals"

Both correct, both self-described.

## Emails: collected, deliberately not committed

The issue asks for contributor emails and specifies 纯爬虫，不要发邮件问. Nothing here sends mail.

On the addresses themselves I made a judgement call worth flagging: `git log` publishes an author's email and most contributors do not realise it, so a committed list of them is a spam and doxxing vector whatever the intent. So:

- emails are gathered only behind an explicit `--author-emails` flag
- they are written to a **gitignored** contacts file, never into the survey
- `noreply` addresses are dropped, since they identify nobody
- a test asserts no harvested address can reach the shareable report

The survey itself carries public professional signals only and is safe to commit or share. If you want the addresses committed too, that is a one-line change to `.gitignore`, but I would not do it by default.

## Verification

- 601 tests pass (596 on main + 5 new), `ruff check` and `ruff format --check` clean
- Ran live against the GitHub API; the 403s on unauthenticated runs surface as warnings rather than silently dropping repositories

Closes #156

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,14 @@ out/
 # store before publishing; the MVP's null extractor makes no external calls.
 /data/kw_bench_classifications.jsonl
 
+# Commit emails harvested by `benchmark-radar authors --author-emails` (issue
+# #156). `git log` publishes an author's address and most contributors do not
+# realise it, so a committed list of them is a spam and doxxing vector whatever
+# the intent. The survey itself carries public professional signals only and is
+# safe to share; this file is not.
+/out/benchmark-author-contacts.json
+benchmark-author-contacts.json
+
 .gstack/
 .DS_Store
 .worktrees/
--- a/src/benchmark_radar/authors.py
+++ b/src/benchmark_radar/authors.py
@@ -1,0 +1,319 @@
+"""Who builds the popular benchmarks, and which of them work on data.
+
+Issue #156 asks which authors behind popular benchmarks are doing data work.
+This answers it from public professional signals only: the GitHub repositories
+the radar already tracks, their contributor lists, and the profile fields those
+contributors chose to publish (name, company, bio, blog, location).
+
+Two deliberate limits.
+
+**Popularity is derived, not asserted.** The issue leaves "popular" to us, so it
+is defined here as a repository the radar's own corpus ranks: stars, how many
+days the radar has tracked it, and how many connectors reported it. That keeps
+the seed list reproducible and auditable rather than a hand-picked list of
+famous names.
+
+**Commit emails are collected but never committed.** `git log` publishes an
+author email, and most contributors do not realise it. A committed file of
+harvested addresses is a spam and doxxing vector whatever the intent, so emails
+go to a separate gitignored artifact and the shareable profile records carry
+professional signals only. `noreply` addresses are dropped entirely; they
+identify nobody and are pure noise.
+
+No email is ever sent. This reads public APIs and stops there.
+"""
+
+from __future__ import annotations
+
+import os
+import re
+from typing import Any
+
+from .corpus import build_corpus
+from .http import RequestError, get_json
+
+AUTHORS_SCHEMA_VERSION = 1
+GITHUB_API = "https://api.github.com"
+DEFAULT_REPO_LIMIT = 40
+DEFAULT_CONTRIBUTORS_PER_REPO = 20
+
+# Substrings that mark a profile as data-adjacent work. Matched against the
+# self-written bio and company fields, never inferred from a repository name:
+# an author is credited with data work because they said so.
+DATA_SIGNALS = (
+    "data quality",
+    "data-centric",
+    "data curation",
+    "dataset",
+    "data engineering",
+    "annotation",
+    "labeling",
+    "labelling",
+    "evaluation",
+    "eval",
+    "benchmark",
+    "rlhf",
+    "human feedback",
+    "data pipeline",
+    "etl",
+    "data infrastructure",
+)
+
+_GITHUB_REPO = re.compile(r"^https?://github\.com/([^/]+)/([^/?#]+)", re.IGNORECASE)
+
+
+class AuthorLookupError(RuntimeError):
+    """A required GitHub call failed and the result would be misleading."""
+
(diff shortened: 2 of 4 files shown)


ktwu01:
Post-merge contact audit completed on 2026-08-08.

The authenticated sweep covered 40 repositories / 378 contributor profiles with zero API failures. It produced 156 non-noreply commit contacts and 4 profile-keyword candidates. I manually audited those four against current profiles, project repositories, and official organization/project materials.

Result: the four matches are not four data vendors.

- Chris Holmes / Planet + Fields of the World: vendor-side, data-relevant lead; building open agricultural field-boundary benchmark/global prediction data plus models and inference tooling. The classifier's `etl` reason was itself a substring false positive inside `PlanetLabs`, so this retention is based on manual evidence.
- Quentin Lhoest / Hugging Face: commercial data-platform/infrastructure lead; building and maintaining Hugging Face Datasets and related dataset tooling. Not verified as a bespoke dataset supplier.
- Jan Batzner / EvalEval: non-vendor research-infrastructure partner; building Every Eval Ever's common schema and community evaluation-results database.
- Andrew Tran / EvalEval: non-vendor secondary contributor/contact working on evaluation infrastructure.

The audited, outreach-ready list (including contact provenance, outreach angle, caveats, and sources) is in the private repo at [benchmark-radar/issue-156](https://github.com/ktwu01/vendor-data-qc/tree/main/benchmark-radar/issue-156), commit [`3b61aab`](https://github.com/ktwu01/vendor-data-qc/commit/3b61aab). No addresses are reproduced in this public comment. The raw 156-address harvest was deliberately not committed; only four manually audited, relevant contacts are retained privately. No email was sent.

PR #201, opened 2026-08-15
docs: simplify README around core use cases

ktwu01:
## Summary

Simplify the README so the first screen answers only the main questions:

- why Benchmark Radar exists
- where to open the GUI
- where to subscribe via RSS
- where to download the curated dataset
- how to contribute

Less important material is moved into a compact `More` section with links to the scoring rubric, model-card data, corpus schema, configuration, local usage, support, contact, and license.

This removes the long methodology-first presentation from the README and makes the project usable within a few seconds of opening the repo.


ktwu01 (commented):
excellent

Issue #241, opened 2026-08-18
recurring: Try to avoid technical jargon. Try to let 16-year-olds read it.


ktwu01:
页面信息量太大了
不知道关注什么
就比如说这一整个page 我不知道哪里是有用的信息
然后他们的字体颜色 weight 之类的还都一样

---

## Scope, after audit (2026-08-19)

This issue was raising two problems at once. They are now separated.

**This issue keeps the jargon half.** It is the recurring parent that the weekly audit refs: #276 is the first weekly report under the workflow added in `17cd2dd`, and a new one arrives every Monday. Nothing to do here except keep serving as the standing rationale.

**The visual-hierarchy half moved to #280.** The lines about 字体颜色 weight 之类的还都一样 and 不知道关注什么 are a typography problem, not a vocabulary problem, and had no owner. Rewriting `corpus` to "everything collected" does not help a page rendered at a single type weight.

**Page named.** The complaint was about `/?view=leaderboard`. The homepage (`/benchmark-radar/`) version of it was handled in #248, closed. The leaderboard additionally needs the split described in #256 (leaderboard itself vs. a separate saturation/progression page); #280 is blocked on that landing first, since moving secondary content off the page reduces what has to be ranked.

Recurring by design: do not close.

### Done when

Never. This is a standing policy issue, not a task.

It exists to give the weekly jargon audit a rationale to point at: #276 and every Monday report after it ref this issue. If it is ever closed, those reports lose their reason for existing.

The two things that would legitimately end it: the weekly workflow is retired, or the policy is superseded by a written style guide that lives somewhere better than an issue body.

**Do not close as part of backlog cleanup.**

ktwu01:
---

## Scope, after audit (2026-08-19)

This issue was raising two problems at once. They are now separated.

**This issue keeps the jargon half.** It is the recurring parent that the weekly audit refs: #276 is the first weekly report under the workflow added in `17cd2dd`, and a new one arrives every Monday. Nothing to do here except keep serving as the standing rationale.

**The visual-hierarchy half moved to #280.** The lines about 字体颜色 weight 之类的还都一样 and 不知道关注什么 are a typography problem, not a vocabulary problem, and had no owner. Rewriting `corpus` to "everything collected" does not help a page rendered at a single type weight.

[...]

PR #271, opened 2026-08-19
Check user-facing text for project vocabulary every Monday

ktwu01:
Refs #241. Does **not** close it: #241 is a standing quality bar, and the point of this is that it keeps applying.

## What this adds

`scripts/audit_jargon.py` scans the strings a reader actually sees (the i18n keys in `app.js`, the copy in the HTML) for 15 terms that are precise inside this project and opaque outside it. A workflow runs it Mondays 02:00 UTC.

Deliberately **not** a readability score. Flesch-Kincaid rewards short words, and "run", "row" and "shard" are short. The failure #241 describes is unexplained vocabulary, not long sentences.

## What it finds today

25 hits: `readable score` (5), `connector` (4), `protocol` (4), `frontier` (4), `instrument` (3), `corpus` (3), `saturation` (2).

Sample of what a reader currently sees:

- "Which of that movement is corroborated by more than one **connector**?"
- "**Corpus** coverage"
- "Two snapshots are available, but their **connector** coverage or report limit differs..."

## Design choices worth reviewing

- **One rolling issue, not one per week.** A new issue every Monday buries the discussion on the previous one. The workflow edits the open `jargon`-labelled issue if there is one, and the title carries the date so staleness is visible.
- **It never closes anything**, per your instruction that fixes get manually verified.
- **Lookup keys and DOM attributes are filtered out**, with a test pinning it. A report containing `data-frontier-point` teaches the reader to skim, and then the real hits go unread too.
- **The term list is hand-written** and lives at the top of the script, so adding or retiring a word is a one-line edit.

## Not done here

The 25 hits are reported, not fixed. Rewriting them is a copy decision per string, and some are fine as-is where the surrounding sentence defines the term. Worth its own PR once you have looked at the list.

836 tests pass; ruff clean.

### DIFF
--- a/.github/workflows/weekly-jargon.yml
+++ b/.github/workflows/weekly-jargon.yml
@@ -1,0 +1,47 @@
+name: Weekly Jargon Audit
+
+on:
+  schedule:
+    # Mondays 02:00 UTC = 10:00 Asia/Singapore, an hour after the daily radar
+    # so the two never contend for the same runner. Weekly rather than daily
+    # because the copy changes slowly and a report nobody has had time to act
+    # on is a report that gets skimmed (issue #241).
+    - cron: "0 2 * * 1"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  audit:
+    permissions:
+      contents: read
+      issues: write
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: "3.12"
+      - name: Scan user-facing text for project vocabulary
+        run: python scripts/audit_jargon.py --markdown > body.md
+      - name: Open or update the tracking issue
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          title="Jargon in user-facing text — week of $(date -u +%Y-%m-%d)"
+          # One rolling issue, not one per week: the finding list is a
+          # standing state of the copy, so a new issue every Monday would
+          # bury the discussion that happened on the last one. The title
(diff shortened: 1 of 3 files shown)


ktwu01:
Audited before merge, rebased onto current main, and found two faults in my own work.

## Conflict

`tests/test_site.py` conflicted because #274 and this PR both appended tests to the end of the file. Both sides kept, nothing dropped: `test_source_mix_names_the_sources_that_found_nothing` and `test_source_mix_separates_the_three_reasons_a_source_shows_zero` from #274, plus `test_jargon_audit_reads_only_user_facing_text` from here. 839 tests pass.

## Two faults found and fixed

**1. An em dash in the issue title.** `"Jargon in user-facing text — week of ..."` would have been stamped onto an issue every Monday forever, against this project's own style rule. Also removed from the separator between each finding and its location.

**2. A clean week still filed an issue.** The script printed "No flagged terms this week. Nothing to do." and the workflow opened or updated an issue saying so. That is the exact failure this PR is supposed to prevent: a recurring notification about nothing teaches you to skip the notification, and then the week with real findings gets skipped too.

Now the script prints nothing when the copy is clean and the workflow exits before touching the tracker. An already-open issue is left alone rather than edited to say it is empty, so work in progress on the list is not lost mid-week.

Verified both paths rather than assuming: a fixture with plain copy exits 0 with zero bytes and files nothing; the real tree emits 3,255 bytes and would file.

## Coverage check

Confirmed the scanner reaches all 394 user-facing strings, including the ones #274 added after this branch was written. It correctly clears them: "found nothing at all" and "could not be reached" are already plain English, which is the point of that PR. No blind spot.

## The 25 findings

Reviewed every one; they are real UI text, not false positives. Examples: "Corpus coverage", "Which of that movement is corroborated by more than one connector?", "No corpus entities yet."

Worth noting "Pareto frontier" is legitimate in its context, which is why the report says in as many words that not every hit is a bug and a term the surrounding sentence defines is fine. This is a triage list, not a build gate.

## Safety

`contents: read` at the top level, `issues: write` only on the job. The `jargon` label it queries exists. It never closes anything.

Issue #257, opened 2026-08-19
update the issue template: make a new field to be VISUAL


ktwu01:
UI is soooo important while UI too bad

ktwu01:
Fixed in #275, merged to main.

The template picker asked people to classify their problem before describing it, and the two forms on offer ("data error", "add a model card") did not cover the case you raised: someone who sees the UI looking wrong. So the honest path was to skip the forms and file blank, which is exactly how UI reports arrived with no screenshot, no page, and no screen size.

There is now **one form**, with the kind picked from a dropdown inside it:

- Something looks wrong (layout, spacing, colours)
- A logo is wrong, missing, or too small to read
- A number looks wrong
- Wording is confusing or too technical
- Something is missing that should be there
- A button or link does not work
- The page is slow or does not load
- Something else

Then where you saw it, then what happened, with the prompt saying a screenshot can be pasted straight into the box. Three fields, and the reporter never has to know how this project files things.

The screenshot is **not required**. A required attachment locks out anyone on a restricted machine or a phone, and a described bug beats a lost one.

`data-error.yml` was retired, since "a number looks wrong" is now one dropdown choice. The model card form stays as a second entry because it is not a bug report: it collects six specific fields that become a data change, and carries `good first issue` as an on-ramp. Blank issues remain enabled for ideas, questions, and proposals.

Also added a `logo` label, and rewrote the remaining form and the `data-error`/`model-card` label descriptions in plain English (refs #241): "mapped to registry ids" and "the counted unit is the document" are gone.

PR #272, opened 2026-08-19
Ask for a screenshot when something looks wrong

ktwu01:
Refs #257

UI quality matters a lot here, and bug reports about the UI keep arriving with no picture. These forms make a screenshot the normal thing to include.

## What was already there

`.github/ISSUE_TEMPLATE/` had `add-model-card.yml`, `data-error.yml`, and `config.yml`. There was no bug form at all. Anyone reporting that the site looked broken was filing a blank issue, which is exactly the path that produces a report with no screenshot, no page address, and no screen size. That gap is most of the problem.

## Separate template, or a field on a generic bug form

Both, and deliberately.

A visual bug and a logic bug need different things. For a visual bug the picture is close to the whole report: what "wrong" means is often impossible to write down but obvious in one image, and it also depends on the screen it was seen on. For a logic bug, the steps are load-bearing and a screenshot is supporting evidence.

Folding both into one form means either asking everyone for screen size and viewport details they do not need, or burying the visual questions so far down that the people who most need them skip them. GitHub's template picker is cheap: a reporter who thinks "this looks wrong" picks the form that says that. So the visual report gets its own form with the screenshot box first, and the general bug form carries the same box lower down, optional.

## Files

Added `ui-bug.yml`, "Report something that looks wrong". Screenshot or recording box first, then page address, then which screen they were on (a plain dropdown, not device widths), then what looks wrong, then browser as optional.

Added `bug.yml`, the general bug form. Points visual problems at the form built for them, and still offers the screenshot box.

Added `logo.yml`, wired to the new `logo` label, which had nothing routing reports to it. Readability at chart size is the failure the label describes, so it asks for the logo shown at the size it actually appears rather than zoomed in.

Changed `data-error.yml`, adding an optional screenshot of the published row. The source quote is still the part that decides the issue, so this stays at the bottom and stays optional.

## Choices worth flagging

The screenshot field is never required. A required attachment is a wall: someone on a locked-down machine, or reporting from a phone, simply cannot file, and a described bug beats a lost one. The forms lean on wording instead, asking for the picture first and saying plainly that it is the fastest route to a fix.

Every form asks for the page address and screen, because those are the two facts that make a visual bug reproducible and the two people most often forget.

Wording follows #241, plain English throughout: "Screen you were on" rather than viewport, "Page address" rather than URL. Each form is five or six fields, since a long form suppresses reports.

Labels are all existing ones: `bug` for the two bug forms, `logo` for the logo form.

## Checks

Every file parses under `yaml.safe_load`, and each form has `name`, `description`, and `body`, with `type` and `attributes` on every body item and `options` present on every dropdown.

`uv run pytest -q`: 835 passed. That matches the count on `origin/main` exactly, so nothing regressed. (The 836 figure in the brief came from a commit on another local branch that adds one test and is not part of this base.)

https://claude.ai/code/session_01PuehBQFtLfJ81UWGv6EGCg

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,0 +1,56 @@
+name: Report a bug
+description: Something on the site does not work. For data that disagrees with its source, use "Report a data error" instead.
+title: "Bug: "
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        If the problem is with how something looks (a chart, a logo, a label, a
+        layout), please use "Report something that looks wrong" instead. It asks
+        for the few extra details that make a visual problem fixable.
+
+  - type: textarea
+    id: what_happened
+    attributes:
+      label: What happened
+      description: What you did, and what the site did instead of what you expected.
+      placeholder: |
+        I clicked the source filter and picked one source. The table went empty.
+        I expected it to show the rows from that source.
+    validations:
+      required: true
+
+  - type: input
+    id: page
+    attributes:
+      label: Page address
+      description: Copy the address bar from the page where it happened.
+      placeholder: https://ktwu01.github.io/benchmark-radar/
+    validations:
+      required: true
+
+  - type: dropdown
+    id: viewport
+    attributes:
+      label: Screen you were on
+      options:
+        - Desktop or laptop
+        - Phone
+        - Tablet
+        - Not sure
+    validations:
+      required: true
+
+  - type: textarea
+    id: visual
+    attributes:
+      label: Screenshot or screen recording
+      description: |
+        Strongly wanted, and usually the fastest way to get this fixed. Drag the
+        image or video file straight into this box and GitHub will upload it.
+        Please crop out anything private first. If you cannot take one, leave
+        this empty and file anyway.
+      placeholder: Drop your screenshot here.
+    validations:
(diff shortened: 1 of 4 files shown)


ktwu01:
Superseded by #275. The ask turned out to be the opposite of more forms: one form, with the kind picked from a dropdown inside it, plus where and what.

#275 keeps what was learned here. From this PR: the finding that no bug form existed at all, which is why UI reports arrived with no screenshot, and the screenshot prompt that fixes it (still never required, so a phone user is not locked out). From #273: the plain-English wording on the model card form.

Suggest closing this one unmerged.

ktwu01:
Closing unmerged, superseded by #275 which is now on main.

This went the opposite way from what was wanted: four forms instead of one. But the research here was the thing that made #275 right, and it is worth recording. This PR found that `.github/ISSUE_TEMPLATE/` had **no bug form at all**, only `add-model-card.yml` and `data-error.yml`. Anyone reporting that the site looked broken was filing a blank issue, which is precisely why those reports arrived with no screenshot, no page, and no screen size. Adding a visual field to an existing form would not have fixed that, because the form the reporter needed did not exist.

#275 carries that forward: the screenshot prompt, and the decision to keep it optional so a phone user is not locked out. The dropdown covers the same cases these three separate forms did, without making the reporter choose between them first.

Nothing here is lost, and the branch is left in place.

Issue #260, opened 2026-08-19
Source mix in /?view=trends shall show those GAPS (source that equal to zeros)


ktwu01:
#254  就是说有些source每天只找到0个新消息。比如说 first hand feed rss
we wanna show a highlight/warning in  /?view=trends to show that "有些source每天只找到0个新消息"

ktwu01:
Fixed in #274, merged to main.

The Source mix column now names every source that ended a day at zero, instead of leaving it out. The important part is that a zero has **three** meanings, not one, and the fix distinguishes them:

- **could not be reached** (the fetch failed) and **found nothing at all** (the fetch worked, zero records) render as red chips, with a note that a source stuck at zero is usually broken rather than quiet
- **nothing ranked** (records arrived, none scored high enough to be listed) renders as a quiet grey chip and raises no alarm

That distinction is not cosmetic. Source mix counts ranked evidence after scoring, while fetch health counts raw records before it, so treating every zero as "found nothing" would report healthy sources as broken. I checked the current data independently: aggregated per source across all 27 days, **every** zero is the harmless `nothing ranked` kind. Without the split, the page would have cried wolf on every one of them.

A sentence above the ledger names the newest day's gaps, since "3 sources are at zero" does not tell you where to look.

Works retroactively across all 27 stored days: the states are derived on the page from data every snapshot already carries, so no pipeline change was needed. Verified in a browser in both English and Chinese, no console errors.

Issue #254, opened 2026-08-19
Data availability: First-party feed and GitHub Release have 0 items on Aug 18, 2026


ktwu01:
## Problem

On August 18, 2026, the source filter `/?source=First-party+feed` and `/?source=GitHub+Release` show no results (empty evidence items).

## Root Cause

These sources genuinely had 0 evidence items ingested on Aug 18, 2026:

| Source | Items on Aug 18 | Notes |
|---|---|---|
| `First-party feed` | 0 | RSS keyword filter (`benchmark`, `evaluation`, etc.) matched 0 of 13 feeds |
| `GitHub Release` | 0 | GitHub API returned no new releases for configured repos |

For comparison, on Aug 13 there was 1 First-party feed item, and on Aug 14 there was 1 each of First-party feed and GitHub Release.

## Impact

- Users clicking `/?source=First-party+feed` or `/?source=GitHub+Release` on Aug 18 see an empty results page
- The filter itself is working correctly — it's a data availability issue, not a filter bug

## Related

- Issue #248: User reports `/?source=First-party+feed` and `/?source=GitHub+Release` have no visual result after the filter on Aug 18, 2026
- Fix applied in this session: Made source filter case-insensitive in `site/assets/app.js` to handle URL parameter casing and `+` encoding consistently

## Resolution Path

This is expected behavior — these are non-guaranteed feeds/APIs that may have zero results on certain days. Consider documenting this as a known limitation or adding a "no results" message when a source filter yields empty results.

ktwu01:
The resolution path this issue proposed is now on main, via #274 (which closed #260).

You wrote: *"Consider documenting this as a known limitation or adding a 'no results' message when a source filter yields empty results."* The Trends ledger now names every source that ended a day at zero, rather than omitting it, so a quiet source is visible instead of looking like it does not exist.

It also separates the case this issue is actually about. First-party feed on Aug 18 shows as **found nothing at all** (a red chip: the fetch worked, zero records matched), which is different from **could not be reached** and different again from **nothing ranked** (records arrived but none scored high enough). Your table drew exactly that distinction in prose, and the page now draws it visually.

Leaving this open rather than closing it, since you did not name it: the remaining half of your suggestion, a message on the filtered Today view itself when a source filter yields nothing, is not done. #274 covers the Trends ledger only. Say the word and I will close it, or do the filtered-view message as a follow-up.

ktwu01:
Both halves are now on main. Closing.

[...]

PR #274, opened 2026-08-19
Show the sources that found nothing, and say which kind of nothing

ktwu01:
Refs #260. Refs #254.

## The problem

On `/?view=trends`, the Daily ledger's **Source mix** column listed only the sources that returned something. A source that came back with nothing simply was not in the row, so a reader could not tell these two apart:

- this source looked and found nothing today
- this source does not exist

That distinction is the whole point. Issue #254 recorded First-party feed and GitHub Release sitting at 0 on Aug 18, 2026, and a source that quietly stays at zero for several days is usually a broken source, not a quiet one. Nothing on the page said so.

## What changed

**The zeros are printed.** Each ledger row now names the sources that ended the day at zero, as small labelled chips after the counts. Fetch health records a run under its internal key (`first_party_feeds`) while the source mix counts finished evidence under the reader-facing label (`First-party feed`), so a small mapping bridges the two.

**A zero has three meanings, and only two are a problem.** The source mix counts ranked evidence after scoring; fetch health counts raw records before it. The ledger's own note already says a source "can be ok and still empty". So the reason a source is at zero is read from both signals:

| state | meaning | shown as |
|---|---|---|
| unreachable | the fetch failed, nothing could arrive | red chip |
| empty | the fetch worked and returned no records | red chip |
| unranked | records arrived, none scored high enough to list | grey chip |

Calling the third one "found nothing" would be wrong. GitHub returning 300 records that all score too low is the ranking working as intended, and dressing that as a fault teaches the reader to skim past the real gaps. On the current local data, 13 of 115 zeros are of that third kind.

**One sentence above the table** names the newest day's gaps in plain words, since the ledger is long and "3 sources are at zero" tells a reader there is a problem but not where to look. Each of the three reasons calls for different follow-up. The "usually broken, not quiet" warning and the red left border are withheld when the only zeros are unranked ones, so a normal day does not cry wolf.

Two smaller fixes that came out of review:

- A source reported by more than one fetch-health row now takes the failure rather than whichever row happened to come first.
- A row that goes on to name its zeros no longer also claims "none".

## User-facing wording

No internal vocabulary. The UI says "source", never "connector". Every new English string has a Chinese entry in the i18n table, and the two sentence templates carry `{date}` and `{sources}` so Chinese can order them its own way.

## Checks

- `uv run pytest -q`: **838 passed** (835 on `main` plus 3 new tests). One of the new tests walks `SOURCE_FETCHERS` in `sources.py` rather than a hand-kept list, so adding a tenth source without a reader-facing name fails here instead of printing a bare internal key in the ledger.
- `uv run ruff check .` and `uv run ruff format --check .`: unchanged from `main` (the same 4 pre-existing errors and 5 pre-existing format diffs, all in files this PR does not touch).
- Browser check at `/?view=trends`, English and Chinese, no console errors. Verified the red and grey chips render on every row, the summary sentence names the right sources for the newest day, and the `is-warning` border appears only when a real gap exists.

## Not done

- No Python or pipeline change. The zeros are derived on the page from data the snapshots already carry, so this works on every historical snapshot rather than only on days generated after the change.
- The `Fetch health` column is left alone. It already reports failures and empty runs; this PR is about the source mix not showing them.

https://claude.ai/code/session_01PuehBQFtLfJ81UWGv6EGCg

### DIFF
--- a/site/assets/app.js
+++ b/site/assets/app.js
@@ -29,6 +29,46 @@ const LEGACY_SOURCE_COLLECTION_METHODS = {
   brave: "API",
 };
 
+// Fetch health records a run under its internal key ("first_party_feeds"),
+// while the source mix counts finished evidence under the label a reader sees
+// ("First-party feed"). Without this bridge a source that returned nothing is
+// simply missing from the mix, and a reader cannot tell "this source looked
+// and found nothing" apart from "this source does not exist" (issue #260).
+const SOURCE_DISPLAY_NAMES = {
+  arxiv: "arXiv",
+  huggingface: "Hugging Face",
+  github: "GitHub",
+  openreview: "OpenReview",
+  semantic_scholar: "Semantic Scholar",
+  github_releases: "GitHub Release",
+  first_party_feeds: "First-party feed",
+  openalex: "OpenAlex",
+  brave: "Brave Web",
+};
+
+const sourceDisplayName = (key) =>
+  SOURCE_DISPLAY_NAMES[key] || String(key).replaceAll("_", " ");
+
+// Sources that ran for this day but put nothing into the ranked evidence, kept
+// in the order fetch health reports them so the ledger reads the same way every
+// day. A failed fetch is already reported by the Fetch health column, so this
+// covers both: a source that errored and one that succeeded with nothing to
+// show are equally invisible in the source mix, and equally worth flagging
+// because a silent zero can mean a broken source (issue #254).
+function zeroItemSources(day) {
+  const counts = day.source_counts || {};
+  const seen = new Set();
+  return (day.ingest_health || [])
+    .filter((entry) => entry.kind !== "attention")
+    .map((entry) => ({ ...entry, name: sourceDisplayName(entry.source) }))
+    .filter((entry) => {
+      if (Number(counts[entry.name] || 0) > 0) return false;
+      if (seen.has(entry.name)) return false;
+      seen.add(entry.name);
+      return true;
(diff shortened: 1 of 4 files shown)


ktwu01:
Verified independently before review, not taken on trust.

**The three-state distinction is right, and matters more than the PR claims.** I recomputed the zeros from `radar.json` myself, aggregating per source rather than per producer row: across 27 days, every single zero is `unranked`, meaning the source fetched records that did not score high enough to be listed. Not one is a genuine fetch failure. Labelling those "found nothing" would have cried wolf about sources that are working correctly, on every occurrence.

(My count differs from the PR's 13-of-115 because I collapse multiple producers per source; the mechanism is the same and the page's per-row granularity is the right one to display.)

**Browser check on this branch**, served locally at `/?view=trends`:
- 115 gap chips, 13 with `is-unranked`, matching the PR
- `browse console --errors` clean
- The red/grey split is legible: `First-party feed 0` and `Brave Web 0` read as problems, `arXiv 0` and `GitHub Release 0` read as quiet
- The caption states it plainly: "Source mix counts ranked evidence after scoring. Fetch health counts raw records returned before scoring, so a source can be ok and still empty."

One note for whoever merges: the PR reports 838 tests against a 835 baseline, which I confirm (the 836 figure in my brief to the agent was off by one, my error). The pre-existing ruff failures it mentions are also real and on `main` already, in files this PR does not touch.

Not merging: leaving that to you per your review-first instruction.

PR #426, opened 2026-08-29
[12 points] Add local-first benchmark query CLI, managed data sync, and Agent Skill

junjiezhou1122:
## What this adds

- Add one shared `QueryService` for deterministic local benchmark search, detail lookup, recent Radar evidence, and data health.
- Expose the same JSON contract through `search`, `show`, `recent`, `status`, and the optional local `serve` HTTP interface.
- Add consumer-facing `init` and `sync` backed by a complete checksummed Pages data bundle.
- Store verified data under the cross-platform `.benchmark-radar` user directory, activate updates atomically, retain only the active version, and make deferred cleanup observable/retryable.
- Add the purpose-neutral `skills/benchmark-radar` Agent Skill, which routes user requests through consumer CLI commands without assuming their research goal.
- Document consumer versus maintainer workflows in English and Chinese, and keep CI/AGENTS guidance aligned.

## Search and update behavior

- Search remains offline, reproducible lexical/token retrieval in this version; this PR does not add semantic search or MCP.
- Each response reports retrieval mode and local data provenance, including the content-sensitive `data_version`.
- Sync checks the small manifest first, supports ETag/304, validates transport, size, SHA-256, ZIP safety, catalog integrity, shard completeness, and manifest counts before activation.
- Failed activation preserves the previous verified version. Windows-style post-activation file locks return `cleanup_pending` and retry before the next network sync.
- `normalize-external`, `classify`, and `build-data-release` remain maintainer/CI commands.

## Validation

- `ruff check .`
- `ruff format --check .`
- `benchmark-radar normalize-external`
- `benchmark-radar classify`
- `benchmark-radar build-data-release`
- `pytest -q` — **1021 passed** after merging current upstream `main`
- Fresh Python 3.12 installed-package smoke: `init`, `sync`, `search`, `show`, `recent`, `status`, and `serve` all exercised end to end against a locally served release; `/healthz` and `/api/v1/search` returned 200 with expected results.
- Skill validator passed; Vercel Skills CLI discovered and installed `benchmark-radar` in an isolated directory.

## Deployment note

The downloader intentionally rejects HTTPS-to-HTTP redirects. The current public Pages/custom-domain routing still redirects the canonical Pages URL to an HTTP custom-domain URL, and `benchmark-radar.org` currently has a certificate mismatch. The Pages workflow in this PR publishes the CLI bundle, but the default public `init` path will remain safely blocked until the HTTPS/custom-domain routing is corrected.


ktwu01:
Merged, thank you @junjiezhou1122. Verified live after deploy: `init` pulls the
5.5 MB archive from the new `cli-data` Release with a checksum matching the
published manifest, the manifest itself is 548 bytes on Pages, and a repeat
`sync` on current data finishes in 0.44s without re-downloading anything. That
is exactly the bandwidth behavior this was aiming for.

Two notes on what changed after your last push:

I added one commit before merging (478150e). `gh release create` needed
`--latest=false`, because GitHub picks the Latest release by creation date, so
the first deploy would have moved the repository sidebar and `/releases/latest`
off v0.9.0 and onto the rolling data drop. Confirmed after deploy that
`cli-data` is `latest=false` and v0.9.0 is still Latest.

The deployment note in the description is now stale in a good way:
`ktwu01.github.io` currently redirects to HTTPS `benchmark-radar.org` with a
valid certificate, so the default `init` path is not blocked. That is why the
end to end run above works against the public URLs.

One follow-up from exercising the merged build: #432. Lexical search fills all
20 slots for `"protein structure prediction"` even though zero records in the
catalog contain `protein`, ranking LiveCodeBench first because its description
mentions "test output prediction". It matters mainly because the Skill routes
agent questions through `search`, so an agent will report that as real coverage.
No rush, and no expectation that you pick it up. Flagging it to you first since
you designed the retrieval path.

Issue #432, opened 2026-08-29
[2 points] search: lexical mode returns confident but irrelevant results when the catalog has no match


ktwu01:
## What happens

`benchmark-radar search` returns a full page of confident-looking results for a
query that has no relevant match anywhere in the catalog.

```
$ benchmark-radar search "protein structure prediction"
20 of 20 matches (lexical, scope=catalog)
  1. LiveCodeBench  [catalog]  llm-stats-livecodebench
     1 of 3 query tokens matched; fields=description
  2. AI2D  [catalog]  llm-stats-ai2d
     1 of 3 query tokens matched; fields=description
  3. CBNSL  [catalog]  llm-stats-cbnsl
     1 of 3 query tokens matched; fields=description
```

LiveCodeBench is ranked first because its description happens to contain the
phrase "test output prediction". It has nothing to do with protein structure.

## Why these results are wrong

Measured against the shipped catalog (1173 records, `data_version`
`2026-08-29T11-24-46.465972Z-c72a736f88c8`):

| token | records containing it |
| --- | --- |
| `protein` | **0** |
| `structure` | 33 |
| `prediction` | 12 |

There is no correct answer to this query in the corpus. The honest response is
zero results. Instead every one of the 20 slots is filled by a record that
matched a single generic word.

The behavior is also inconsistent. `search "weather forecasting"` correctly
returns `0 of 0 matches`, because neither token appears anywhere. So whether a
user gets an honest empty result or a page of noise depends on whether one
common English word happens to show up in some unrelated description.

Queries where the catalog does have real coverage work well today, which is
worth preserving:

```
$ benchmark-radar search "math reasoning"
20 of 480 matches (lexical, scope=catalog)
  1. MATH       all query tokens matched across fields; fields=name,description,categories
  2. We-Math    all query tokens matched across fields; fields=name,description,categories
  3. MATH-500   all query tokens matched across fields; fields=name,description,categories
```

## Why this matters more than a normal ranking nit

#426 ships `skills/benchmark-radar`, which routes user questions through these
commands. An agent has no way to tell a 1-of-3 generic token match from a real
hit, so it will faithfully report LiveCodeBench as a protein structure
prediction benchmark. A wrong answer delivered confidently is worse than an
empty one, and Benchmark Radar's value rests on being trustworthy about what it
does and does not cover.

The `--json` contract makes this hard for a caller to defend against. Each
result carries `rank` and a human-readable `match` string, but no numeric
[...]

PR #459, opened 2026-08-30
fix: make lexical retrieval evidence agent-judged

junjiezhou1122:
## Related issue

Related to #432

## Hypothesis

Lexical search should retrieve inspectable evidence; the consuming Agent/Skill should decide whether a benchmark satisfies the user. A service-side semantic acceptance rule cannot reliably distinguish conversational wrappers (`find`, `please`, `I need`) from domain terms without introducing brittle false positives and false negatives.

## Baseline

Issue #432 showed `protein structure prediction` filling the default page with ordinary-looking one-token matches. The previous response exposed prose and rank but not a clear retrieval score, IDF coverage, or missing-token evidence. Conversely, treating every raw user token as mandatory falsely downgraded `find GPQA Diamond` and `SWE bench verified please`.

## Intervention

- Keep deterministic BM25F retrieval with bounded name/phrase boosts and no query-specific branches.
- Rename the query-local ranking field to `retrieval_score`; expose smoothed `idf_coverage` alongside matched/missing tokens, fields, phrases, and score components. These are retrieval signals, never confidence.
- Keep `full_matches_found`, `partial_candidates_only`, and `no_lexical_candidates` as lexical retrieval statuses. Partial candidates remain available for Agent inspection.
- Update the consumer Skill to strip conversational wrappers, issue two to four focused variants, compare candidates, inspect `show` details, and make the final `verified match` / `possible candidate` / `no confident match` judgment.
- Add wrapper-language cases (`find GPQA Diamond`, `SWE bench verified please`, `I need a medical VQA benchmark`) to the qualitative judge set; it now contains 23 cases total.
- Preserve both OpenCompass round-1 card provenance and round-2 enrichment provenance in normalized records.

## State diff

No semantic search, MCP, remote Search API, stopword list, or query-specific exception was added. `README.md`, `README.zh-CN.md`, and their tests remain identical to current upstream `main`. The service contract is shared by CLI and HTTP.

## Metric result

The deterministic 18-query local evaluation remains green:

- navigational Hit@1: 1.00
- Hit@5: 1.00
- MRR@20: 1.00
- macro Recall@20: 1.00
- catalog-gap full-match rate: 0.00
- reviewed partial retention: 1.00

Representative rebuilt-catalog cases:

| Query | Retrieval status | Evidence |
| --- | --- | --- |
| `GPQA Diamond` | `full_matches_found` | exact name first, IDF coverage 1.0 |
| `protein structure prediction` | `partial_candidates_only` | 0 full, 29 partial; missing token evidence exposed |
| `weather forecasting` | `no_lexical_candidates` | zero lexical overlap |
| `math reasoning` | `full_matches_found` | `MATH` first |

## Guardrail result

Clean Python 3.12 worktree passed the canonical six-step CI sequence and `pytest -q` (1,077 passed). Focused retrieval, metadata, evaluation-schema, and Skill tests: 97 passed. Local evaluator passed. Independent spec review found no findings; independent standards review found and we fixed one unused token-cache smell.

## Ablation

An all-token service gate would hide useful partial evidence and reject natural-language wrappers. Keeping retrieval high-recall while delegating acceptance to the Agent preserves evidence without claiming semantic suitability.

## Decision requested

Please review the updated retrieval-only contract and Skill-owned acceptance boundary. If #432 must require a service-side semantic floor, that should be specified separately from this retrieval-layer change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,4 +40,7 @@ jobs:
       - run: benchmark-radar normalize-external
       - run: benchmark-radar classify
       - run: benchmark-radar build-data-release
+      # Ranking evaluation consumes the generated Catalog and fails on a
+      # versioned retrieval-quality threshold before the broader test suite.
+      - run: python scripts/evaluate_search.py
       - run: pytest -q
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,4 @@ jobs:
       - run: benchmark-radar normalize-external
       - run: benchmark-radar classify
       - run: benchmark-radar build-data-release
-      # Ranking evaluation consumes the generated Catalog and fails on a
-      # versioned retrieval-quality threshold before the broader test suite.
-      - run: python scripts/evaluate_search.py
       - run: pytest -q
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -177,14 +177,16 @@ upload that exact file. The PDF is a dated interpretation, not a data source.
   errors; they must not be replaced with guessed metadata.
 - Lexical search is a high-recall candidate retriever for agents, not a final
   suitability judge. Any shared query token may produce a candidate. BM25F is
-  the primary score. Exact/prefix/token-sequence name matches and non-name
-  contiguous phrases are bounded, query-IDF-scaled boosts; weighted coverage is
-  only a tie-breaker and explanation because BM25F already rewards additional
-  matched terms. Never require every query term as an eligibility gate. Every
-  result must expose matched and missing tokens, fields, coverage, and score
-  components. `full_matches_found` means at least one candidate covers every
-  query token, `partial_candidates_only` preserves weaker evidence without
-  claiming an answer, and zero token overlap uses `no_lexical_candidates`.
+  the primary retrieval score. Exact/prefix/token-sequence name matches and
+  non-name contiguous phrases are bounded, query-IDF-scaled boosts; lexical
+  coverage is only a tie-breaker and explanation because BM25F already rewards
+  additional matched terms. Every result must expose matched and missing tokens,
+  fields, coverage, `retrieval_score`, `idf_coverage`, and score components.
+  `full_matches_found` means at least one candidate covers every query token,
+  `partial_candidates_only` preserves weaker evidence without claiming an answer,
+  and zero token overlap uses `no_lexical_candidates`. Semantic acceptance belongs
+  to the consuming Agent/Skill, which may issue focused query variants and inspect
+  `show` details before making a suitability claim.
 - Catalog and Radar are different trust layers. Catalog rows are normalized
   benchmark records; Radar rows are discovery evidence and must stay labelled as
   such. A search result is a candidate, not a suitability claim. Agent query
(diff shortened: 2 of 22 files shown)


ktwu01 (changes requested):
Thank you — this is a meaningful ranking and transparency improvement. I reproduced the cleaner BM25F ordering, the structured match evidence, and the explicit partial-only terminal message. I also ran the canonical clean-checkout sequence successfully: all 1,077 tests passed.

Before merging, please improve the acceptance/status contract:

1. The linked bug is not fully fixed yet. `protein structure prediction` still returns 29 candidates and displays the first 20 by default. They are now honestly labelled partial, which fixes the confident presentation and gives agents useful evidence, but the service still has no relevance floor or `no_confident_match` decision. One shared token remains sufficient and `candidate_count == total_matches`. Please either add an actual service-side abstention/acceptance rule or narrow the claim so this PR does not close #432 as specified.

2. The current all-query-token status creates false abstentions for normal phrasing. In the rebuilt catalog:
   - `find GPQA Diamond` ranks GPQA Diamond first but returns `partial_candidates_only` because `find` is missing.
   - `I need a medical VQA benchmark` ranks PMC-VQA first but returns `partial_candidates_only` because `i` and `need` are missing.
   - `SWE bench verified please` ranks SWE-Bench Verified first but is partial because `please` is missing.

   This conflicts with the Skill guidance that partial-only output must never be treated as an answer. Please add natural-language and exact-name-with-wrapper regression cases, then handle non-discriminative wrapper tokens and/or recognize a strong contained name match without letting generic one-token description overlap become confident again.

3. Please avoid locking the disputed solution into `AGENTS.md` as “Never require every query term as an eligibility gate” until the acceptance policy is settled. The repository instruction should describe the agreed contract, not preclude the relevance floor requested by #432.

4. The OpenCompass metadata restoration is useful and correctly surfaces StructTokenBench. Because descriptions, categories, and modality now come from the round-1 snapshot while record provenance still names only the round-2 bundle, please preserve both input provenances in the normalized record.

The ranking implementation and evaluation scaffolding are good foundations. The main remaining requirement is a status/acceptance boundary that rejects weak evidence without falsely rejecting obvious named or naturally phrased searches.

junjiezhou1122:
Updated after review and latest upstream rebase (`main@a64db57`). I kept the acceptance boundary in the consuming Agent/Skill rather than adding a global stopword list or semantic service-side floor: the service cannot know whether tokens such as `find`, `please`, or `need` are wrappers or domain requirements. The service now exposes `retrieval_score` and smoothed `idf_coverage`, while `search_status` is explicitly lexical retrieval status. The Skill now removes conversational wrappers, tries focused variants, inspects `show`, and can make the final `no confident match` judgment. Added three wrapper-language qualitative cases and preserved both OpenCompass input provenances. I changed the PR relationship to `Related to #432`; if the issue requires service-side semantic rejection, that is a separate policy layer and should be specified independently.

junjiezhou1122:
Rebased once more onto the latest upstream `main@42f3d66` (the intervening commits only update the site privacy disclosure). The clean Python 3.12 six-step CI was rerun from this exact head: `ruff check`, formatting, all three generators, and `pytest -q` all passed (`1,079 passed`).

PR #521, opened 2026-09-02
Add a daily brief blog at /blog/

ktwu01:
## What this adds

A daily brief blog at `/blog/`, built from the committed snapshots. One page per
collection day, plus a landing page (latest 30), a full archive, and its own RSS
feed at `/blog/feed.xml`.

- `/blog/` and `/blog/archive/` list every brief.
- `/blog/YYYY-MM-DD/` is one day's report.
- `/blog/feed.xml` is a separate feed. The existing `/feed.xml` is unchanged.

The dashboard gains exactly one thing: a `Blog` link in the menubar after
Explore. It carries no `data-view`, so it is a normal page load and not an SPA
route. No existing view, control, or `styles.css` rule changes.

## Three kinds of day, each labelled

- **Daily brief** for a snapshot with a stored briefing: bullets, questions,
  cited statistics, confidence, counter-views, caveats, and provenance.
- **No material change** for a `no_material_insight` briefing, published
  unchanged. "Nothing moved enough to call it a change" is that day's report,
  and dropping it would leave a hole in the record.
- **Evidence summary** for the early snapshots from before briefings were
  stored. These show the day's counts and its first ten evidence records, and
  say on the page that they are a deterministic summary, not a synthesized
  briefing.

Nothing here calls a model or the network. Every word comes from the snapshot,
so `classify` stays safe in CI and a rebuild is byte-identical.

## Reader-facing details

- Citations show the stored `E###` and `S###` IDs. The prose cites by ID and the
  IDs are not contiguous, so a page that printed only titles could say `E011`
  while showing that document as item 3.
- Chinese is published when the snapshot stored a reviewed translation, with a
  toggle. Pages without one ship no toggle and no invented translation. Fixed
  question prompts reuse the dashboard's translations, and a test fails if the
  two tables disagree.
- Falling back to English on an untranslated brief does not overwrite a reader's
  stored language preference, which is shared with the dashboard.
- Each brief gets its own sitemap `lastmod`, so a deploy does not ask crawlers to
  refetch the whole archive.

## Verification

- Full CI sequence green on a clean `git worktree` checkout: ruff check, ruff
  format, normalize-external, classify, build-data-release, 1177 tests passing.
- Every route smoke-tested from the built site: `/blog/`, `/blog/archive/`,
  `/blog/feed.xml`, and per-day pages of all three kinds.
- Dashboard compared before and after at 1440x900 and 390x844. The only visible
  difference is the new menubar item.

`site/blog/` is generated and gitignored, like the rest of `site/data/`.

Please merge with a merge commit, not a squash, per the repository policy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,6 +82,11 @@ graphify-out/
 # reason as the benchmark pages: the generator and inputs are reviewed, not a
 # stale HTML copy.
 /site/leaderboard/
+# One page per collection day, written by the same `classify` run from the
+# committed snapshots. The whole tree is replaced on every build, so a
+# committed copy would be a stale brief reviewed as if it were the day's
+# report, and a day removed from the history would keep its page.
+/site/blog/
 /site/trends/
 /site/explore/
 /site/cli/
--- a/site/assets/app.js
+++ b/site/assets/app.js
@@ -383,6 +383,7 @@ const I18N = {
     Leaderboard: "排行榜",
     Trends: "趋势",
     Explore: "探索",
+    Blog: "博客",
     Rubric: "评分标准",
     "Daily briefing": "每日简报",
     "Questions for today": "今日问答",
--- a/site/assets/blog.css
+++ b/site/assets/blog.css
@@ -138,9 +138,22 @@
   color: var(--muted);
 }
 
-.blog-page .blog-list,
-.blog-page .blog-evidence {
+.blog-page .blog-list {
   padding-left: 1.25rem;
+}
+
+/* The source list is keyed by stored citation ID, not by position, so it
+   carries no marker of its own. The ID sits in its own column so a title that
+   wraps stays aligned under itself rather than under the reference. */
+.blog-page .blog-evidence {
+  padding-left: 0;
+  list-style: none;
+}
+
+.blog-page .blog-evidence li {
+  display: grid;
+  grid-template-columns: 3.2rem 1fr;
+  align-items: baseline;
 }
 
 .blog-page .blog-list li + li,
--- a/site/assets/blog.css
+++ b/site/assets/blog.css
@@ -1,0 +1,286 @@
+/* Blog pages use the dashboard's stylesheet as their visual source of truth
+   and only compose its existing tokens into a reading layout. Nothing here
+   applies to the dashboard: every rule is scoped to .blog-page, which only the
+   generated blog documents carry. */
+.blog-page .masthead-end {
+  display: flex;
+}
+
+.blog-page .blog-view {
+  padding: clamp(2.25rem, 4vw, 3.75rem) 0 6rem;
+}
+
+.blog-page .blog-hero {
+  display: grid;
+  gap: 1rem;
+  max-width: 980px;
+  margin-bottom: 2rem;
+}
+
+.blog-page .blog-hero h1 {
+  max-width: 920px;
+}
+
+.blog-page .blog-lede {
+  max-width: 760px;
+  margin: 0;
+  color: var(--muted);
+  font-size: clamp(1rem, 2vw, 1.25rem);
+}
+
+.blog-page .blog-actions,
+.blog-page .blog-tags {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 0.65rem;
+  align-items: center;
+}
+
+.blog-page .blog-stats {
+  display: grid;
+  grid-template-columns: repeat(3, minmax(0, 1fr));
+  gap: 1rem;
+  margin: 0 0 2.5rem;
+}
+
+.blog-page .blog-stat,
+.blog-page .blog-panel,
+.blog-page .blog-card {
+  border: 1px solid var(--edge);
+  border-radius: var(--radius);
+  background: var(--panel);
+  box-shadow: var(--shadow);
+}
+
+.blog-page .blog-stat {
+  padding: 1.1rem;
+}
+
+.blog-page .blog-stat strong {
+  display: block;
+  font-family: var(--display);
+  font-size: clamp(1.35rem, 3vw, 2rem);
+  line-height: 1.1;
+}
+
+.blog-page .blog-stat span,
+.blog-page .blog-meta,
+.blog-page .blog-kicker {
+  color: var(--muted);
+  font-family: var(--utility);
+  font-size: 0.7rem;
+  letter-spacing: 0.05em;
+}
+
+.blog-page .blog-panel {
+  margin: 1.5rem 0;
+  padding: clamp(1rem, 2.5vw, 1.6rem);
+}
+
+.blog-page .blog-panel > :last-child,
+.blog-page .blog-card > :last-child {
+  margin-bottom: 0;
+}
+
+.blog-page .blog-section {
+  margin: 2.5rem 0;
+}
+
+.blog-page .blog-section-heading {
+  margin-bottom: 1rem;
+  padding-bottom: 0.75rem;
+  border-bottom: 1px solid var(--edge);
+}
+
+.blog-page .blog-section-heading h2,
+.blog-page .blog-section-heading p {
+  margin-bottom: 0;
+}
+
+.blog-page .blog-grid {
+  display: grid;
+  grid-template-columns: repeat(3, minmax(0, 1fr));
+  gap: 1rem;
+}
+
+.blog-page .blog-card {
+  padding: 1.1rem;
+}
+
+.blog-page .blog-card h2,
+.blog-page .blog-card h3 {
+  margin-bottom: 0.45rem;
+}
+
+.blog-page .blog-card p {
+  color: var(--muted);
+}
+
+.blog-page .blog-table {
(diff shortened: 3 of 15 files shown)


ktwu01:
@junjiezhou1122 兄弟 你方便在你的本地看看效果怎么样吗？

比如说，这个PR 有没有把原来的 好的 UI 改掉换成不好的

junjiezhou1122:
> @junjiezhou1122 兄弟 你方便在你的本地看看效果怎么样吗？
> 
> 比如说，这个PR 有没有把原来的 好的 UI 改掉换成不好的

ok~

junjiezhou1122:
## Local verification (checked out `feat/daily-brief-blog`, clean rebuild)

Everything in the PR description checks out. Evidence from the local run:

**CI sequence**
- `ruff check .` ✅ · `ruff format --check .` ✅
- Full `pytest -q`: **1177 passed** — matches the claimed count
- `benchmark-radar classify` regenerates `site/blog/` without errors

**Rebuild determinism** — verified, not just claimed:

```
find site/blog -type f | sort | xargs md5 -q   # run classify again
→ diff: empty (byte-identical)
```

**All three kinds of day confirmed in the generated output** (41 day pages):
- 28 daily briefs, 12 of which carry the "No material change" report — published unchanged, as designed
- 13 evidence summaries from pre-briefing snapshots, explicitly labelled "deterministic summary" on the page

**Behavior spot-checks in a real browser**
- Language toggle switches the whole document (`<html lang>` flips to `zh-CN`), preference is persisted under `benchmark-radar:lang`, and the toggle correctly does not appear on untranslated pages
- Untranslated brief keeps the English title (no invented translation) while showing the stored `*_zh` blocks — matches the "publish only what was reviewed" rule
- Sitemap: each brief carries its own `lastmod` (e.g. `/blog/2026-08-18/` → `2026-08-18`), not the site-wide one
- Dashboard: the only change is the new `<a href="/blog/">` with no `data-view`; `APP_VIEWS` untouched, so briefs arrive as full page loads
- `/blog/feed.xml` is a separate feed; the root `/feed.xml` is unchanged

**One data-layer observation (not a blog bug, no action required on this PR)**

In the Chinese rendering of recent briefs, phrases like `Why it matters:` and `Evidence: E001` stay English inside otherwise Chinese sentences. Checked the source: `data/snapshots/*/briefing.bullets_zh` is stored that way — the blog publishes the reviewed translation faithfully rather than inventing one, which is exactly the documented behavior. If the mixed-language look bothers anyone, the fix belongs in the snapshot briefing generator, not here. Worth a follow-up issue if the zh quality matters for the audience.

Nice work — the "no material change is still a report" and "evidence summaries never claim to be briefs" distinctions are the right call for an honest archive. Merge-commit ready per repo policy.

junjiezhou1122:
UI我自己也测试看了下，没发现什么问题，我让ai也测试了一遍，也没有啥问题！ 只是我点击blog的时候好像是空的目前～ 然后我从 http://localhost:8734/blog/ 为啥我从这个界面点击 today 跳转到benchmark-radar.org那里 我继续看下

junjiezhou1122:
**Follow-up fix pushed (`dfd2450`): blog navigation stays on the serving host**

Found while clicking through the local preview: the shell's brand link and all five section links were absolute `SITE_URL` URLs, so every nav click on `localhost` (or any non-canonical mirror) bounced to `https://benchmark-radar.org`. The dashboard's own menubar already uses root-relative paths for the same routes, so the shell now does too.

- `<a class="brand" href="/">` and `navigation()` now emit root-relative `href`s
- `SITE_URL` is kept where an absolute URL is semantically required: canonical, `og:*`, sitemap locs, feed links, schema.org
- Re-verified after the change: ruff clean, **1177 passed**, blog rebuild still byte-identical across two `classify` runs, and clicking Today from a locally served brief now stays on the local host

One clarification for anyone testing before merge: `https://benchmark-radar.org/blog/` 404s right now because the blog is generated, not committed — `site/blog/` is produced by `classify` in the fresh Pages checkout. The URL goes live when this PR merges and Pages redeploys; `pages.yml` already runs the generator that writes it.

junjiezhou1122:
image
image
blog 界面和其他界面有点不太一样，例如右上角？

junjiezhou1122:
**Root-cause fix pushed (`9306244`): the blog chrome is now extracted from `site/index.html`, not restated**

The follow-up that added Rubric/CLI to `_NAV_LINKS` treated a symptom. The real problem: the blog shell hand-copied the site's masthead, nav, and footer, and the copy had already drifted on every axis —

- no Rubric/CLI tabs, no Contact button, no Star/Fork/Issues badge group, a bare footer, and a different RSS target;
- the Rubric/CLI tabs that were added rendered as plain links, because `styles.css` styles the dashboard's dialog-trigger tabs **by id** (`#rubric-nav`/`#cli-nav` get the inactive blue and the "ⓘ" marker), and the shell didn't carry those ids;
- the language toggle had invented its own `aria-label`/`data-next-language` contract instead of the dashboard's title/glyph/aria-pressed one.

`site/index.html` is the single hand-maintained source of that chrome — every dashboard route is it served at its own path — so `write_blog` now **extracts the masthead, nav, and footer from it at build time** and applies a small, explicit transform set, each justified by the blog being a static document outside the SPA:

- Today's view button becomes the plain link to `/`; `data-view`/`aria-controls`/`aria-expanded` are dropped, **ids stay** (they're CSS hooks);
- Contact becomes a link to `/#contact` — the dashboard deep link that opens the same sheet on load (verified: loads with the dialog open);
- the masthead RSS badge keeps the site feed with a root-relative href;
- the toggle ships only on translated pages, driven with app.js's exact contract;
- the footer build date is baked from the day the page describes.

Consequence: a future menubar/badge/footer edit in `index.html` propagates to every brief automatically, and a parity test (`test_blog_nav_lists_the_same_sections_in_the_same_order_as_the_dashboard`) fails loudly if the extractor falls behind the dashboard's shape.

Re-verified after the change: ruff clean, **1182 passed**, blog rebuild byte-identical, and in a real browser the blog's nav, badges, contact deep link, toggle, and footer now render and behave identically to the dashboard's. (Badge counts stay "—" when api.github.com rate-limits an unauthenticated client — same graceful fallback the dashboard has.)

junjiezhou1122:
**Fix pushed (`2fe7d1e`): the chrome now translates for Chinese readers, from the same reviewed table**

After the chrome extraction, the remaining top-right gap was language: the dashboard translates its chrome in place from app.js's `I18N.zh` table, and the blog — which deliberately does not load app.js — kept an English masthead. Same split for hover tooltips: on the dashboard the badges show immediate CSS tooltips (`data-tooltip`), on a brief there was nothing.

- The build now parses `I18N.zh` out of app.js (quoted **and** bare object keys — `Contact: "联系作者"` has no quotes) and bakes the subset the extracted chrome needs into every page as `<script type="application/json" id="chrome-i18n">`.
- `blog.js` applies it with the exact `applyI18n` contract: `data-i18n` → text, `data-i18n-title` on a repo-badge → `data-tooltip` + title removed + aria-label fallback, `data-i18n-aria` → aria-label; switching back to English restores the originals.
- Keys the table doesn't carry stay English — mirroring the dashboard's own `t()` fallback rather than inventing translations (nav labels like Today are translated, so the brief's tabs read 博客/命令行 there too).
- The footer's `Updated` prefix and the contact link's `data-i18n-title` join the table; badge count aria-labels reuse the `{count}` templates.

Verified in a browser, both states, side by side with the dashboard: EN — native titles on RSS/Contact; ZH — `联系作者`, `切换到英文` on the toggle, `通过 RSS 订阅 Benchmark Radar` on hover. Plus ruff clean, **1184 passed**, and a byte-identical blog rebuild.

junjiezhou1122:
**Fix pushed (`c381500`): the top right is now pixel-identical — proven with a pixel diff**

Two last deltas, both caught by rendering both pages and diffing the pixels instead of eyeballing markup:

1. **The contact bubble was a filled blob on briefs.** Its svg carries no class; on the dashboard the outline comes from `button.repo-badge svg`, a rule the rebuilt link could never match. The extractor now tags that svg with the `outline-icon` class the fork/issues icons already use on link badges.
2. **The landing and archive pages had no language toggle at all** (the old "toggle only where a translation exists" rule) — so `/blog/`, the page the menubar link opens, was missing the toggle the dashboard always shows. Since the chrome is now translatable on every page, the toggle is never inert: it ships everywhere, and a body without a stored translation simply stays English.

Result, from a pixel diff of the rendered chrome zone (masthead + nav row) between the dashboard and `/blog/`: **zero differing pixels in the top right**. The only remaining difference anywhere in the chrome is each page's own active nav tab — Today highlighted on the dashboard, Blog on a brief — which is the "you are here" state working as intended.

ruff clean, **1184 passed**, blog rebuild byte-identical.

ktwu01:
兄弟。你觉得 ready 的话，merge 就可以啦

junjiezhou1122:
image
the rubric 消失

PR #595, opened 2026-09-10
Describe the score sources accurately in both READMEs

ktwu01:
The acknowledgement paragraph said the frontier-model score layer "is built on benchmark data collected by LLM Stats." That was inaccurate on two counts, both checked against the repo's own data.

**LLM Stats is one of three sources, not the basis.** `site/data/models.json` source counts are Artificial Analysis 594, LLM Stats 339, model reports 53.

**The SWE-bench Verified timeline is not LLM Stats data.** The README GIF links to `/saturation/?lfrontier=swe_bench_verified`, which resolves to the record whose `source` is `model_reports` (lab system cards). A separate `llm-stats-swe-bench-verified` record exists, but it is not the timeline the sentence pointed at.

## Changes

Both READMEs now say the scores draw on three kinds of source, name lab model reports as the SWE-bench timeline's actual source, credit Artificial Analysis and LLM Stats for nearly all remaining model scores, and add OpenCompass Hub as the wider catalog's fourth source. Line 51 already named all four; the acknowledgement had dropped OpenCompass.

## On the LLM Stats license

Their [terms](https://llm-stats.com/legal/terms-of-service) are more permissive than the old wording implied: reuse and modification are allowed, including commercially, with attribution as the only condition. Credit must be "in a place visible to your users, with a link back to llm-stats.com."

That condition is not currently met on the website itself, where no page links back. A follow-up PR adds the credit to the `/cite/` page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,9 +163,13 @@ and [Hacker News](https://news.ycombinator.com), plus first-party lab feeds from
 [IBM Research](https://research.ibm.com), [Databricks](https://www.databricks.com),
 [LangChain](https://www.langchain.com/blog), and [Meituan Engineering](https://tech.meituan.com).
 
-The frontier-model score layer, including the SWE-bench Verified timeline above,
-is built on benchmark data collected by [LLM Stats](https://llm-stats.com).
-Thank you for keeping that data open.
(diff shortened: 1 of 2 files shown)


ktwu01 (commented):
okay

PR #597, opened 2026-09-10
Credit the score sources on the citation page

ktwu01:
LLM Stats grants reuse, including commercial use, on one condition: credit `llm-stats.com` "in a place visible to your users, with a link back" ([terms](https://llm-stats.com/legal/terms-of-service)). The only link lived in the README, and no page on benchmark-radar.org linked back. The citation card is where a reader goes to ask where this data came from, so the credit closes that card.

## What it looks like

A single muted sentence, last in the panel, after "View the citation file":

> Benchmark score data comes from lab model reports and from LLM Stats, Artificial Analysis and OpenCompass Hub.

Deliberately quiet by request: no heading, no card, no border, no button styling. It is a footnote under the formats a reader came for, not a section competing with them. Lab model reports are named without a link because they are publications rather than one site, but they are named, since they supply scores too.

## Both renderers, one sentence

`site/cite/index.html` is generated and gitignored, and the panel is rendered twice: the Python seed in `app_seeds.py` is what crawlers and script-disabled readers get, and `openCite()` in `app.js` overwrites it on first paint. A credit in only one of them is a credit half the readers never see, so both draw from one source list and a test asserts they agree.

## Verification

Six-step CI sequence on a clean detached worktree, not the working copy, since gitignored build artifacts on disk mask failures CI hits:

- `ruff check` and `ruff format --check` pass
- `normalize-catalog`, `classify`, `build-data-release` all run
- **1329 passed** on the clean checkout

Three tests fail in my local working copy (`test_catalog_coverage`, `test_models_registry`, `test_score_filters`), but they fail identically on unmodified `main` and all pass on the clean checkout. Stale generated artifacts, not this change.

Also confirmed:

- `llm-stats.com` is present in the generated no-JS seed, which is what crawlers receive
- Served the built site and checked the panel in a browser: credit renders last, exactly once (the JS render replaces the seed rather than duplicating it), links inline and clickable
- The copy-button count is still 3, so `test_seeded_copy_controls_...` is unaffected
- Both new `t()` keys resolve in the zh table; `and` reuses the existing entry

New tests cover the compliance link, seed/JS agreement, and the zh translations, since nothing previously caught an untranslated string.

### DIFF
--- a/site/assets/app.js
+++ b/site/assets/app.js
@@ -983,6 +983,8 @@ const I18N = {
     "Copy it with your keyboard": "请用键盘复制",
     "Citation file (.cff)": "引用文件 (.cff)",
     "View the citation file": "查看引用文件",
+    "Benchmark score data comes from lab model reports and from":
+      "benchmark 分数数据来自各家实验室的模型报告，以及",
     CLI: "命令行",
     "This search covers all dates.": "此处搜索全部日期的结果。",
     "Still want today's results?": "仍要搜索今天的，请",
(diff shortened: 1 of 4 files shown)


ktwu01:
Dropped an unintended submodule pointer bump (`docs/technical-report/latex`) that had ridden along in the first commit. Advancing that pointer is a separate decision and does not belong in a citation-card change.

The diff is now four files: `app.js`, `styles.css`, `app_seeds.py`, `tests/test_app_pages.py`.

ktwu01 (commented):
okay

PR #604, opened 2026-09-11
Make the Cite card reader-first and add CITATION.md (#464)

### LINKED ISSUE #464: [2 Points] Make the "cite" button in the bottom-left of this home page to be more user friendly.
- [x] #566 
- [x] pending arXiv submission to get the arXiv link!

Currently, the site button is redirecting to this link, but this link is not user-friendly at all.
https://github.com/ktwu01/benchmark-radar/blob/main/CITATION.cff

below can be wrong-- but, Basically, we want the user to only cite this technical report. That's it. No need to cite other things, so we should not redirect to this link.

- [ ] Also / Instead, we should make a citation.md page, which contains:
```md
you can use [xxx] (plain APA format) to cite this tech report
if you are AI agent, you can read https://github.com/ktwu01/benchmark-radar/blob/main/CITATION.cff
```

- [ ] also as for https://github.com/ktwu01/benchmark-radar/blob/main/CITATION.cff, we only keep one thing to cite, which is the tech report on arXiv

ktwu01:
Closes #464.

The footer Cite button already opens the /cite/ card instead of the raw CITATION.cff link, but the card gave the machine-readable file equal billing with the APA and BibTeX a reader actually needs.

What changes:
- New CITATION.md: leads with copy-paste APA and BibTeX for the arXiv tech report (arXiv:2609.11115, the only thing to cite), links the paper and the /cite/ page, and reserves CITATION.cff for AI agents.
- Cite card (site dialog plus server-rendered seed, both languages): the third copy block and its link are relabeled as AI-agent-only. APA and BibTeX stay first.
- Both READMEs show the APA first, state only the tech report needs citing, and point at CITATION.md. CITATION.cff already cites only the tech report via preferred-citation, so it is unchanged.
- New regression test pins CITATION.md to citation.apa_citation() and the agent note; updated the dialog-label assertion in test_site.py.

Test plan:
- ruff check . and ruff format --check . pass.
- pytest tests/test_readme.py tests/test_site.py tests/test_query_surfaces.py tests/test_app_pages.py tests/test_technical_report.py: 217 passed.
- Full pytest -q after normalize-catalog, classify, build-data-release: 1329 passed, 1 failed in tests/test_i18n_benchmark_term.py (strict ji-zhun check). That failure is pre-existing and unrelated: it fires only on docs/technical-report/latex submodule content (the local checkout pins 8e64d7c vs 951ea96), reproduces with this change stashed, and none of the hits are in files this PR touches.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -7,6 +7,16 @@ authors:
     given-names: Koutian
   - family-names: Zhou
     given-names: Junjie
+  - family-names: Shang
+    given-names: Ergan
+  - family-names: Wang
+    given-names: Jiayu
+  - family-names: Han
+    given-names: Pengqian
+  - family-names: Wang
+    given-names: Junkai
+  - family-names: Xu
+    given-names: Wanghan
 repository-code: "https://github.com/ktwu01/benchmark-radar"
 url: "https://benchmark-radar.org/"
 license: MIT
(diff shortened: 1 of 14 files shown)


ktwu01 (commented):
okay. lean and good

PR #609, opened 2026-09-11
Fix the isPartOf structured-data warning and date benchmark pages individually

ktwu01:
Two invisible SEO defects, plus the docs they are recorded in. No UI changes.

## `Invalid object type for field "isPartOf"`

Search Console reports this against the site. The value was a bare
`{"@id": "https://benchmark-radar.org/#website"}`. An `@id` on its own only
resolves on a page that also defines the node it names, and `#website` is
defined on the homepage and its route copies only, so the reference pointed at
nothing on all 1,284 benchmark pages, the benchmark directory, and every blog
post. Blog postings referenced their `publisher` and their parent blog the same
way.

All of them now emit self-describing nodes, so each reference is valid on the
page that carries it.

## Per-benchmark sitemap `lastmod`

Every benchmark URL carried the site-wide snapshot date, so 1,284 of the 1,344
listed URLs changed their `lastmod` on each deploy and invited crawlers to
refetch the whole catalog daily. The date now comes from the page's own
evidence: the newest reported score date, else the record's release date, else
no `lastmod`. Against the published catalog that is 429 distinct dates spanning
2015 to 2026, with 10 pages left undated rather than guessed at.

This mirrors how the blog already dates each brief.

## Docs

The live status table claimed 1,225 sitemap URLs; the published sitemap has
1,344. The node-reference rule is written down so the warning does not come
back.

## Verification

`pytest`: 1,332 passed. `ruff check src/ tests/`: clean. The one failing test on
this branch, `test_no_ji_zhun_in_user_facing_sources_strict`, fails identically
on `main` and is caused by the dirty `docs/technical-report/latex` submodule.

### DIFF
--- a/docs/seo-indexing-guide.md
+++ b/docs/seo-indexing-guide.md
@@ -17,7 +17,7 @@ Last checked: **September 3, 2026**.
 | HTTP redirect | **Passing:** apex HTTP redirects to the HTTPS apex | Keep it |
 | `www` redirect | **Passing:** HTTPS `www` redirects to the HTTPS apex | Keep it |
 | `robots.txt` | **Current:** sitemap uses `benchmark-radar.org` | Keep it |
-| `sitemap.xml` | **Current:** all 1,225 URLs use `benchmark-radar.org` | Keep it |
+| `sitemap.xml` | **Current:** all 1,344 URLs use `benchmark-radar.org`, and each benchmark page carries the date of its own newest evidence | Keep it |
 | Page metadata | **Current:** canonical and `og:url` use the HTTPS apex | Keep it |
 
 Re-run the checks below after any DNS or Pages change. The table is a dated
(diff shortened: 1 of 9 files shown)


ktwu01:
Checked both review points against the published catalog rather than accepting them on votes.

**Taken: the suppressed comment about calendar validity.** Correct. `_ISO_DATE` checked shape, not the calendar, so a shard carrying `2026-02-30` would have published it as a `lastmod` no crawler can parse. Now validated with an exact round trip through `date.fromisoformat`, with a regression test. No value in the published catalog changes (`2f023be`).

**Not taken: using `released` only when no score date exists.** This changes nothing in practice and is worse in principle.

In practice: across all 1,284 shards there is not one where `released` is later than the newest `reported_date`, so the two rules produce identical output on every page the site publishes.

In principle: `lastmod` means the newest fact the page carries, and the page renders the release date in its facts list. Where a release date were the later of the two, reporting an older date than something the reader can see on the page is a regression, not a fix. `max()` over both is the behavior that matches what `lastmod` claims.

Also verified while checking: zero calendar-invalid values and zero future-dated values across the catalog.

PR #612, opened 2026-09-12
SEO: an About page, Publications aliases, and benchmark pages that answer the search

ktwu01:
about #593 (website half).

A reader who finds this site through a search engine now lands on a page that says what it is, and the catalog's own pages answer the words that reader typed.

## What a search result says

The homepage title and heading name the thing the site tracks instead of the product, and the description is the sentence a person reads before deciding to click.

Each benchmark page used to be a stub: a name, a description, and a link back to the dashboard. Someone looking for SWE-bench types the name plus "paper", "dataset", "code", "leaderboard" or "results". The page now answers those from the shard it already had: what the benchmark is, its artifact links, the scores on record, which sources cite it, and sibling benchmarks in the same capability. A section is dropped when the shard holds nothing for it, so a page never promises an artifact the catalog does not have.

## New pages

`/about/` carries the project's story: what the catalog is, who collects it, and where the work is published.

The citation sheet keeps its `/cite/` path and is now labelled **Publications** in the menu, because that is the word a researcher looks for. `/publications/` and `/publication/` are alias documents that canonicalize to `/cite/`, carry `noindex,follow`, and move the reader across. They stay out of the sitemap so the ranking signal stays on one URL.

Both new paths reach the sitemap through their own argument rather than `INDEXABLE_VIEWS`, which stays the set of paths `site/assets/app.js` owns metadata for.

## One fact, said once

The dashboard said several things twice. The banner's words moved into the page that was already below it, the priority score is now a number with a bar and a definition behind `(i)` instead of five restatements of itself, and the information mark is gone from the menu row where it had nothing to explain.

The blog had the same problem in sharper form. A brief printed its kind three times, its date twice, and its opening paragraph twice, once clipped mid-sentence a screen above the full version. The heading is now the whole hero, and an index card is a title and what the day found. The date still goes out in the `BlogPosting` schema and the summary still goes to the meta description and the feed, which is where a clipped summary belongs.

`/about/` and `/blog/` are reachable but not in the visible menu row, which lists the views over the corpus.

## Also in the diff

Shard slugs are held to the slug alphabet before they become a directory name, a canonical URL, and a query argument, none of which escape their interpolation. The dashboard seeds and the static pages now share one table of score-source labels, so a source is named the same way in both places.

`docs/seo-indexing-guide.md` documents the new paths, the aliases, the sitemap boundary, and the post-deploy checks for each.


docs/seo-indexing-guide.md
```diff
@@ -153,6 +154,69 @@ a second URL for a page that already has one is a duplicate, not a second page.
...
 rubric versions use `/rubric/?version=<number>`.
 
+### Pages that are documents, not views
+
+Two more paths carry the same masthead and footer but hold prose rather than a
+control over the corpus:
+
+```text
+/about/
+/blog/
```
ktwu01, line 164: need add `/publication/`, `/publications/` 
these also need SEO

site/assets/app.js
```diff
@@ -975,6 +975,10 @@ const I18N = {
     "If this saved you research time, cite the work, star the repo and help other eval builders find it.":
       "如果它帮你节省了研究时间，请引用这项工作、给仓库点 Star，让更多评测开发者找到它。",
     Cite: "引用",
+    About: "关于",
+    "About this site": "关于本站",
+    Publications: "论文与引用",
+    "Search AI, LLM, agent, and multimodal benchmarks": "搜索 AI、大模型、智能体与多模态 benchmark",
```
ktwu01: add a domain
AI for sci benchmarks

site/assets/app.js
```diff
@@ -1653,9 +1657,9 @@ async function onPopState() {
 // translates through data-i18n.
 const VIEW_SEO = {
   today: {
-    title: "Benchmark Radar: AI Benchmark Tracker & Dataset",
+    title: "AI Benchmark Search Engine & Database | Benchmark Radar",
     description:
-      "A daily evidence-first map of new AI benchmarks, evaluations, and datasets, collected every day from arXiv, GitHub, Hugging Face, OpenReview, Semantic Scholar, Hacker News, and first-party lab feeds.",
+      "Search an open database of AI and LLM benchmarks. Every benchmark gets its own page carrying the paper, code, dataset, leaderboard and scores it has.",
```
ktwu01: add a domain
AI for sci benchmarks

docs/seo-indexing-guide.md
ktwu01: Added, and you were right that the current setup was fighting itself.

The block now lists all four paths with the aliases marked as redirects, plus a link down to their own section:

```text
/about/
/blog/
/publications/  (alias, resolves to /cite/)
/publication/   (alias, resolves to /cite/)
```

On "these also need SEO": checking Google's canonicalization guidance, the stubs were carrying `noindex,follow` **and** a canonical pointing at `/cite/`, and those two pull against each other. Google's wording is that `noindex` "will completely block the page from Search" rather than folding the page into its canonical. So the alias was built to forward its ranking signal to `/cite/`, and the `noindex` was throwing that signal away.

Fixed in 5a024d2: the `noindex` is gone, the canonical stays. An alias is now crawlable and consolidates onto `/cite/` the way Google documents for a duplicate inside one site.

The aliases still stay out of the sitemap, and I would push back on changing that one. A sitemap should list canonical URLs; listing `/publications/` beside `/cite/` asks Google to treat one page as two and splits the ranking signal we just stopped discarding. Crawlable and consolidating gets you the outcome you want: either spelling reaches the page, and `/cite/` is the URL that ranks.

The guide's post-deploy check is updated to match (each alias should now print a canonical and no `robots` line), and I added a Search Console step: after a recrawl, URL Inspection on `/publications/` should report `/cite/` as the Google-selected canonical.

site/assets/app.js
ktwu01: Added in 0824552.

This string is the homepage's visually-hidden `<h1>`, so it is the heading a crawler reads for the whole site. It now runs:

```
Search AI, LLM, agent, multimodal, and AI-for-science benchmarks
```

with the Chinese key updated to `搜索 AI、大模型、智能体、多模态与科学智能 benchmark`.

I checked the catalog before adding the claim rather than just appending the words: it carries 39 benchmarks labelled `AI for Science` and 36 under `Scientific Reasoning`, plus 24 `science`, 15 `physics`, 9 `Medical`, 8 `biology` and 8 `chemistry`. So the heading is describing something the database actually holds.

site/assets/app.js
ktwu01: Added in 0824552, in all four places this description is published.

```
Search an open database of AI, LLM, agent, multimodal and AI-for-science
benchmarks, each with its own page carrying the paper, code, dataset and scores.
```

That is `VIEW_SEO.today.description` plus the three tags in `site/index.html` it has to stay in step with (`meta description`, `og:description`, `twitter:description`), since the homepage's head block is hand-maintained rather than written from `VIEW_SEO`. I also added the domain to `ABOUT_DESCRIPTION`, which enumerates the same list for `/about/`.

It is 153 characters, so it still fits in a search result without truncation. Fitting the new domain in meant trading "Every benchmark gets its own page carrying the paper, code, dataset, leaderboard and scores it has" for the shorter "each with its own page carrying the paper, code, dataset and scores". The word that dropped out is "leaderboard". Say the word if you want it back and I will cut something else instead.

src/benchmark_radar/site_pages.py
```diff
@@ -218,14 +386,91 @@ def _scores_sections(scores_by_source: dict[str, Any]) -> str:
...
+    the dashboard's JavaScript. A page about a benchmark that does not link to
+    the benchmark is the thinnest kind of catalog page there is.
+    """
+    if not artifacts:
+        return ""
+    labels = [_ARTIFACT_LABELS[kind].lower() for kind in _facet_kinds(artifacts)]
+    heading = f"{name} {_join_words(labels).replace(' & ', ' and ')}" if labels else f"{name} links"
+    items = "".join(
+        f'<li><span class="artifact-kind">{_esc(label)}</span> '
+        f'<a href="{_esc(url)}" rel="nofollow noopener">{_esc(url)}</a></li>'
```
toby-bridges, line 410: [P2] Validate the URL scheme before rendering artifact links

HTML escaping does not neutralize a javascript: URL. A constructed OpenCompass identity_evidence.repo_id entry with value="example/bench" and evidence_url="javascript:alert(document.domain)" survives catalog_opencompass._artifacts() and write_benchmark_pages(), producing an executable href in this new section. A reader clicking that link could run script in the site's origin; rel="nofollow noopener" does not prevent this. This is a synthetic reproduction, not evidence of a malicious URL in the production corpus. Please allow only valid HTTP(S) URLs, matching the dashboard's safeHttpUrl behavior, and add a regression test ensuring unsafe schemes never become links.

src/benchmark_radar/site_about.py
```diff
@@ -0,0 +1,278 @@
...
+# the heading directly above it a second time.
+ABOUT_DESCRIPTION = (
+    "An open, daily-updated database and search engine for AI benchmarks: LLM, agent, "
+    "multimodal, coding, safety and AI-for-science evaluations, each with the paper, "
+    "code and scores it has."
+)
+ABOUT_LEDE = (
+    "Every benchmark the catalog has found, with the paper, code, dataset and scores on "
+    "record for it. Collected from public sources and rebuilt every day, so a number here "
+    "is never older than the last run."
```
toby-bridges, line 61: [P2] Distinguish the rebuild date from the dates of the measurements

The claim that a number here is never older than the last run does not hold for the generated catalog. The clean build still correctly displays historical SWE-bench Verified scores, including observations dated 2024-12-27 and 2025-02-24. Rebuilding the site does not remeasure those scores or refresh every source observation. Please describe the catalog as rebuilt daily while retaining each observation's original reported/retrieved date, so readers do not interpret historical measurements as freshly verified values.

toby-bridges (changes requested):
Reviewed commit 5a024d2129fdf68cdac09f373be75d3ea38a9960. Please address the two inline findings before merging: restrict artifact links to HTTP(S), and correct the About page's freshness claim.

The clean rebuild generated all 1,284 catalog pages. The Publications redirect and bilingual navigation worked in the local browser check.

Validation in a fresh detached worktree with the pinned paper submodule initialized: ruff check ., ruff format --check ., benchmark-radar normalize-catalog, benchmark-radar classify, benchmark-radar build-data-release, then pytest -q. All six steps passed; 1,355 tests passed, 0 failed. Local environment: macOS, Python 3.12.13, Node 24.19.0; GitHub CI uses Node 22 and is also green. The existing tests do not cover the unsafe-URL case reproduced below.

Issue #545, opened 2026-09-04
[3 points] add OpenAIRE info source


ktwu01:
Follow-up to #506. Crossref is handled by the pull request that closes #506; this issue tracks OpenAIRE only.

## Goal

Add OpenAIRE as an independent daily-discovery source for benchmark papers, datasets, software, and related research products.

## Acceptance criteria

- Use an official OpenAIRE API and document authentication, rate limits, date semantics, supported product types, and pagination.
- Preserve upstream identifiers, titles, dates, authors, organizations, descriptions, and artifact links without generating source prose.
- Enforce both ends of the collection window, reject malformed or future-dated records, and expose failures through source health.
- Register the optional connector in configuration, scoring/provenance, backfill when supported, and the dashboard source-name map.
- Add tests for valid, empty, malformed, undated, and failed responses, then pass the repository clean-checkout CI sequence.

If production access requires credentials, contact ktwu001 before wiring secrets into the workflow.

PR #613, opened 2026-09-12
Add DataCite as a daily-discovery source

### LINKED ISSUE #544: [3 points] add DataCite info source
Follow-up to #506. Crossref is handled by the pull request that closes #506; this issue tracks DataCite only.

## Goal

Add DataCite as an independent daily-discovery source for DOI-bearing benchmark papers, datasets, and research artifacts.

## Acceptance criteria

- Use the official DataCite API and document authentication, rate limits, date semantics, and pagination.
- Preserve upstream identifiers, titles, dates, authors, organizations, descriptions, links, and available public counters without generating source prose.
- Enforce both ends of the collection window, reject malformed or future-dated records, and expose failures through source health.
- Register the optional connector in configuration, scoring/provenance, backfill when supported, and the dashboard source-name map.
- Add tests for valid, empty, malformed, undated, and failed responses, then pass the repository clean-checkout CI sequence.

If production access requires credentials, contact ktwu001 before wiring secrets into the workflow.

lazizbekravshanov:
Closes #544

## What this gives the radar

- **A second DOI registry in the daily feed.** DataCite registers the DOIs that Zenodo, Figshare, Dryad and institutional repositories mint, so datasets, software and reports deposited there now have their own path into the corpus, beside the Crossref connector that covers publisher DOIs.
- **No credentials, no secrets, no workflow change.** The DataCite REST API is anonymous for reads. `daily-radar.yml` is untouched, so there is nothing to coordinate on credentials.
- **Six requests per daily run** (one page per search phrase) against a limit of 3,000 per 5 minutes per IP.
- **One commit on top of `main` at 5d9e378**, 10 files, 699 additions. Clean-worktree CI sequence passes; details under Verification.

## Two things to decide

1. **The README source count moves from 37 to 38.** That is the figure `count_ingest_sources` in `scripts/export_report_figure_data.py` derives, and `test_readmes_state_the_live_ingest_source_count` now holds both READMEs to it. The repository description on GitHub also says "37 public sources"; that is a setting rather than a file, so it is left for you.
2. **Merge order with the OpenAIRE PR** (#545). Both start from the same `main` commit and both write 38. Whichever lands second conflicts on the two README lines and on adjacent registration lines in `sources.py`, `config.yml`, `pipeline.py`, `rubric.py`, `corpus.py`, `app.js` and the two test files. Every one is an adjacent-insertion conflict, resolved by keeping both sides and writing 39. I will rebase the second one the same day the first merges.

## Acceptance criteria from #544

| Criterion | Where |
| --- | --- |
| Official API; authentication, rate limits, date semantics and pagination documented | `GET https://api.datacite.org/dois`; documented in the `datacite` block of `config.yml` and in the connector docstring |
| Identifiers, titles, dates, authors, organizations, descriptions, links and public counters preserved, no generated prose | `fetch_datacite` carries DOI, titles, `registered`/`published`/`updated`, creators with affiliations, abstract, landing URL, and citation, download and view counts as deposited; no summary is synthesized |
| Both ends of the window enforced; malformed and future-dated records rejected; failures reach source health | `registered:[since TO now+tolerance]` in the query and again on the parsed record; future-dated on either timestamp rejected; unreadable shapes raise `ConnectorPayloadError`, HTTP failures `RequestError`, both surfaced as an optional, non-fatal source |
| Registered in configuration, scoring/provenance, backfill, dashboard source-name map | `config.yml`, `SOURCE_FETCHERS` / `_PARSER_VERSION_METHODS` / `SOURCE_DEFAULT_METHODS`, `EVIDENCE_PRIMARY_SOURCES`, `PRIMARY_SOURCE_RANK`, `BACKFILL_SOURCES`, both maps in `site/assets/app.js` |
| Tests for valid, empty, malformed, undated and failed responses; clean-checkout CI passes | `tests/test_sources.py` and `tests/test_pipeline.py`, listed under Tests; CI under Verification |

Records merge only on the exact DOI, which is the key `dedupe_keys` already uses, so nothing here does fuzzy matching. Nothing is scraped behind authentication.

## How the connector behaves

- **Window and record date are the same field.** The window is `registered`, the moment the DOI started to resolve. `published` is often a bare year and `created` can predate registration by a long draft period, so neither bounds a daily window. A record is dated by that same instant: `updated_at` is the registration timestamp, not DataCite's mutable `updated`, because recency scoring and simulated backfill both read `updated_at`, and keying it on an edit would hand a years-old DOI a new record's recency score and place it in a backfilled day the connector's own window guard rejects. The deposited `updated` stays in `raw`. This matches the Crossref sibling, which dates records by `published`.
- **Only `findable` DOIs are used.** An anonymous read only ever returns that state; the connector skips any other state as a guard for a credentialed run. A deposit registered first and made findable later is therefore collected on its registration date or not at all, and the config comment says so.
- **Search phrases are escaped** before they reach the `query_string` clause. An unescaped colon in a phrase such as `Benchmark: agents` reads as a lookup on a field the non-dynamic mapping lacks and returns zero rows while source health still reports the source healthy. The shipped phrases contain no reserved characters, so the emitted query is unchanged today.
- **Names are preserved as deposited.** Only an explicitly personal name is reordered out of `Family, Given`. `nameType` is optional in the DataCite schema, and assuming a missing one means "personal" would publish `University of California, Berkeley` as `Berkeley University of California`.
- **Titles and abstracts.** An untyped title is preferred over a subtitle, falling back to a typed one rather than dropping a record's only name; a typed `Abstract` is preferred over a technical note.
- **Pagination.** `page[size]` is capped at 1,000 by the API and `page[cursor]` exists for deeper pulls; the connector takes one page per phrase, so a run costs `max_requests` calls. `registered` is not a server-side sort, so results are ordered by `-created` upstream and re-sorted by registration here.
- **Failures raise rather than skip.** A non-object payload, a missing `data` array, or a malformed `attributes`, `titles`, `descriptions`, `creators` or `affiliation` shape is a parsing gap the dashboard should show, not a quietly thinner record.

## Wiring

- `config.yml`: optional `datacite` block (6 phrases, `page_size: 50`, `max_requests: 6`), with the semantics above documented in place.
- `src/benchmark_radar/sources.py`: `fetch_datacite` and its registrations.
- `src/benchmark_radar/pipeline.py`: `BACKFILL_SOURCES`; the registered date range makes a past window honestly re-derivable.
- `src/benchmark_radar/rubric.py`, `src/benchmark_radar/corpus.py`: primary evidence credit and provenance rank 2, beside Crossref.
- `site/assets/app.js`: both source-name maps.
- `README.md`, `README.zh-CN.md`: acknowledgement list and source count.

Zenodo mints its DOIs through DataCite, so the same deposit can arrive from both connectors; both records carry the DOI and merge on it.

## Tests

`tests/test_sources.py`: a valid record (every preserved field and the exact bounded query), undated, out-of-window, future-dated on either timestamp, non-findable state, untitled rows, a title with only a typed entry, name and affiliation shapes, six malformed row shapes, cross-search DOI dedupe with newest-first ordering, limit truncation, a minimal deposit, and query escaping. Plus DataCite entries in the shared empty / malformed / HTTP-failure / no-synthesized-summary parametrizations and the collection-method default.

`tests/test_pipeline.py`: `test_simulate_backfill_places_a_datacite_doi_on_its_registration_day` runs the real connector through `simulate_backfill`, so the connector's window key and the backfill placement key are checked against each other rather than separately.

Each behaviour above that a test protects was checked by reverting it and confirming the test goes red, so the tests carry their regression rather than restating current output.

## Verification

Clean-worktree run of the CI sequence (`git worktree add --detach`, submodule initialized) on this exact head: `ruff check .`, `ruff format --check .`, `normalize-catalog`, `classify` and `build-data-release` pass; `pytest -q` reports 1358 passed, 1 failed.

**One honest caveat.** The live DataCite API was not reachable from the sandbox this was built in, so the request and response shapes were verified against DataCite's own API implementation (lupo: `openapi.yaml`, the `/dois` controller, the Elasticsearch mapping and the JSON:API serializer) rather than a live call. The first daily workflow run after merge is the first live exercise; source health will show any surprise, and I will watch it.

## Not in this PR

Related identifiers stay in `raw`. Carrying `IsSupplementTo` or `IsVersionOf` links as artifact URLs would merge a dataset with its paper, but any cited DOI would merge it with everything it cites, so that needs its own review. The same applies to repositories that register a concept DOI alongside each version DOI: both rows are findable and only a long shared title currently merges them. The connector does not paginate past the first page per phrase, which the config block states.


lazizbekravshanov:
Thanks to both reviewers. Six of the seven findings were real and are fixed in the pushed commit.

The Codex finding on the title query was correct and was the most serious: the config block documented an all-terms rule that the code did not implement, because `query_string` defaults to OR. Terms are now joined with an explicit `AND`, and a test asserts the emitted query contains no `OR`.

Copilot's five findings on shape handling were also correct. `attributes.get(field) or []` looked like a missing-key default but also swallowed a present-but-unreadable `{}`, `""` or `0`, so a row was silently dropped or published without its authors while source health still reported the source healthy. Array fields now default only when the field is absent or null, and a title or description that is not a string raises instead of being stringified into a Python repr.

One finding I do not think holds, and I would rather say so than change the code quietly. On `registered` and `_optional_date`: the concern is that a date-only value is assigned midnight UTC and can then "pass the lower-bound check". Midnight is the earliest instant of that day, so it makes the lower-bound comparison `registered < window_start` strictly harder to pass, not easier; such a record is rejected more often, not admitted. `_optional_date` is also shared by every connector here, and its UTC assumption is deliberate, since calling `astimezone()` on a naive value would move the same record by hours between a laptop and the UTC runner. Special-casing DataCite against that shared helper on this premise would make the code less consistent without fixing a reachable failure. Happy to revisit if you read it differently.

ktwu01 (changes requested):
Thank you for the careful connector work and for addressing the earlier review findings. I verified the DataCite query against the live public API, and the exact merge with current  passes the full local CI sequence with 1,424 tests.\n\nTwo data-boundary cases still need correction before merge:\n\n- , , and  use , so an absent or null upstream measurement is published as a measured zero. Please include only counters that are actually present while preserving explicit zero values; this matches the missing-measurement contract already used when corpus deltas are computed.\n- Creator , , and  values are passed through . A malformed object or array is therefore published as a Python representation, even though title and description scalars correctly raise. Please apply the same string-shape validation to creator names and add regression cases.\n\nThe core connector, date-window behavior, and current-main integration otherwise look good. Thanks for tightening these last two provenance edges.

ktwu01 (commented):

The exact merge with current `main` passes the full local CI sequence with 1,424 tests. Two data-boundary cases remain:

- `citationCount`, `downloadCount`, and `viewCount` use `or 0`, so an absent or null upstream measurement is published as a measured zero. Please include only counters that are actually present while preserving explicit zero values.
- Creator `name`, `givenName`, and `familyName` values are passed through `str(...)`. A malformed object or array is therefore published as a Python representation, even though title and description scalars correctly raise. Please apply the same string-shape validation and add regression cases.

The core connector, date-window behavior, and current-main integration otherwise look good. Thank you.

ktwu01:
Thanks for the explicit-AND query and the absent-vs-malformed array fixes; both
verify in-head with the tests.

1. **Creator name fields go through `str(...)`**, so a malformed object
   publishes as a Python repr while title and description correctly raise.
   Apply the same string-shape validation and add cases. (#614 has the same
   issue on its author and organization fields.)

On the `or 0` counters at `sources.py:1277-1279`: not a blocker here, same
reasoning as on #614. `principle.md` forbids treating unknown counts as zero
and the 19 existing occurrences on `main` do not comply with it, but a
one-connector fix would leave the corpus inconsistent. Filing it repo-wide.

Nits: the merge-order note cites #545, which is the issue; the OpenAIRE PR is
#614. README source count needs updating on rebase, and #632 is a third
claimant on that number.

---

ktwu01:
Filed the repo-wide counter issue referenced above: #676. It lists all 19 call sites across the ten connectors on `main`, and notes that this PR's instances follow the existing convention rather than introducing the problem.

PR #614, opened 2026-09-12
Add OpenAIRE as a daily-discovery source

### LINKED ISSUE #545: [3 points] add OpenAIRE info source

lazizbekravshanov:
Closes #545

## What this gives the radar

- **Research products the radar reaches through no other source.** OpenAIRE aggregates publications, datasets, software and other products from repositories across Europe and beyond, including institutional repositories that neither Crossref nor arXiv covers.
- **No credentials, no secrets, no workflow change.** The Graph API is anonymous for reads. `daily-radar.yml` is untouched, so there is nothing to coordinate on credentials. A registered token raises the budget from 60 to 7,200 requests per hour if it is ever wanted; the connector does not need it.
- **Six requests per daily run** (one page per search phrase) against the anonymous budget of 60 per hour.
- **One commit on top of `main` at 5d9e378**, 10 files, 925 additions. Clean-worktree CI sequence passes; details under Verification.

## Two things to decide

1. **The README source count moves from 37 to 38.** That is the figure `count_ingest_sources` in `scripts/export_report_figure_data.py` derives, and `test_readmes_state_the_live_ingest_source_count` now holds both READMEs to it. The repository description on GitHub also says "37 public sources"; that is a setting rather than a file, so it is left for you.
2. **Merge order with the DataCite PR** (#544). Both start from the same `main` commit and both write 38. Whichever lands second conflicts on the two README lines and on adjacent registration lines in `sources.py`, `config.yml`, `pipeline.py`, `rubric.py`, `corpus.py`, `app.js` and the two test files. Every one is an adjacent-insertion conflict, resolved by keeping both sides and writing 39. I will rebase the second one the same day the first merges.

## Acceptance criteria from #545

| Criterion | Where |
| --- | --- |
| Official API; authentication, rate limits, date semantics, supported product types and pagination documented | `GET https://api.openaire.eu/graph/v3/research-products`, the Graph API v3 that superseded v1 and v2 in July 2026 (the older `/search/*` API was announced for phase-out on 2026-05-31); all five documented in the `openaire` block of `config.yml` and in the connector docstring |
| Identifiers, titles, dates, authors, organizations, descriptions and artifact links preserved, no generated prose | `fetch_openaire` carries the DOI and other PIDs, main title, `publicationDate` and `dateOfCollection`, authors as deposited, organisations by legal name, short name or acronym, description, instance URLs and code repository, plus the citation, download and view counters; no summary is synthesized |
| Both ends of the window enforced; malformed and future-dated records rejected; failures reach source health | `fromPublicationDate` / `toPublicationDate` in the query and again on the parsed record; future-dated rejected; unreadable shapes raise `ConnectorPayloadError`, HTTP failures `RequestError`, both surfaced as an optional, non-fatal source |
| Registered in configuration, scoring/provenance, backfill, dashboard source-name map | `config.yml`, `SOURCE_FETCHERS` / `_PARSER_VERSION_METHODS` / `SOURCE_DEFAULT_METHODS`, `EVIDENCE_PRIMARY_SOURCES`, `PRIMARY_SOURCE_RANK`, `BACKFILL_SOURCES`, both maps in `site/assets/app.js` |
| Tests for valid, empty, malformed, undated and failed responses; clean-checkout CI passes | `tests/test_sources.py` and `tests/test_pipeline.py`, listed under Tests; CI under Verification |

Records merge only on the exact DOI, the key `dedupe_keys` already uses, so nothing here does fuzzy matching. Nothing is scraped behind authentication.

## How the connector behaves

- **Filter values are quoted.** A `mainTitle` value holding whitespace, parentheses or a bare logical operator has to reach v3 double-quoted or the request is answered with HTTP 400 rather than a search. Every phrase this project ships contains a space, so the connector quotes them, escapes an embedded `"`, and leaves a value that is already quoted or parenthesised intact. Without this the source would return nothing at all in production while looking configured.
- **Window and record date are the same field.** `publicationDate` bounds the query on both ends and dates the record. `dateOfCollection` says when OpenAIRE indexed a product rather than when it appeared, and is routinely months later; dating by it would hide a product on its publication day and surface it on a day the connector's own window guard rejects.
- **The DOI is reduced to its bare name.** Repositories deposit it bare, as a link and as a `doi:` URI. A link passed through would build `https://doi.org/https://doi.org/...`, which resolves nowhere and shares no identity with the Crossref or Zenodo copy of the same artifact, so the record would be counted twice.
- **A record needs an openable location.** A product whose only openable location is an instance URL or an http(s) code repository keeps that as its address; one that names no openable location at all is dropped rather than published pointing at nothing.
- **Pagination.** `pageSize` is capped at 100 by the API and the connector clamps a larger configured value; `sortBy=publicationDate DESC` so a page holds the newest matches. One page per phrase, so a run costs `max_requests` calls. A page that matched nothing arrives as `results: null`, which is treated as an ordinary empty result rather than a broken payload.
- **Names are preserved as deposited**, with no reordering; a non-string title raises rather than publishing its repr.
- **Failures raise rather than skip.** A non-object payload, a missing `results` array, or a malformed `authors`, `pids`, `instances` or `organizations` shape is a parsing gap the dashboard should show, not a quietly thinner record.

## Wiring

- `config.yml`: optional `openaire` block (6 phrases, `page_size: 50`, `max_requests: 6`), with the semantics above documented in place.
- `src/benchmark_radar/sources.py`: `fetch_openaire` and its registrations.
- `src/benchmark_radar/pipeline.py`: `BACKFILL_SOURCES`; the publication date range makes a past window honestly re-derivable.
- `src/benchmark_radar/rubric.py`, `src/benchmark_radar/corpus.py`: primary evidence credit and provenance rank 2, beside Crossref.
- `site/assets/app.js`: both source-name maps.
- `README.md`, `README.zh-CN.md`: acknowledgement list and source count.

## Tests

`tests/test_sources.py`: a valid record (every preserved field and the exact bounded, quoted query), undated, out-of-window, future-dated, no openable location, instance and repository fallbacks, a non-openable repository, author and organisation shapes, malformed row shapes, cross-search DOI dedupe with newest-first ordering, limit truncation, the `pageSize` cap, a minimal product, the request budget, quoting of plain, parenthesised, already-quoted and embedded-quote phrases, and the DOI deposited bare, as a link and as a `doi:` URI. Plus OpenAIRE entries in the shared empty / malformed / HTTP-failure / no-synthesized-summary parametrizations and the collection-method default.

`tests/test_pipeline.py`: `test_simulate_backfill_places_an_openaire_product_on_its_publication_day` runs the real connector through `simulate_backfill`, so the connector's window key and the backfill placement key are checked against each other rather than separately.

Each behaviour above that a test protects was checked by reverting it and confirming the test goes red, so the tests carry their regression rather than restating current output.

## Verification

Clean-worktree run of the CI sequence (`git worktree add --detach`, submodule initialized) on this exact head: `ruff check .`, `ruff format --check .`, `normalize-catalog`, `classify` and `build-data-release` pass; `pytest -q` reports 1384 passed, 1 failed.

**One honest caveat.** The live OpenAIRE API was not reachable from the sandbox this was built in, so the request and response shapes were verified against OpenAIRE's own API contract tests plus two independent client libraries that agreed on the response shape, rather than a live call. The first daily workflow run after merge is the first live exercise; source health will show any surprise, and I will watch it.

## Not in this PR

Related identifiers and `relations` stay in `raw`; carrying them as artifact URLs needs its own review, since any cited DOI would merge a record with everything it cites. The connector does not paginate past the first page per phrase, which the config block states.


lazizbekravshanov:
Thanks. One of the three findings was real and is fixed in the pushed commit.

The README source total was genuinely inconsistent: the opening line said 38 while the Abstract still said "37 sources: 13 direct connectors and 24 first-party". Both mentions now read 38 and 14, and `tests/test_readme.py` derives all three figures from `count_ingest_sources` and asserts both mentions, so a future connector cannot update one and leave the other. The same fix is on the DataCite PR, which had the same defect.

The other two I do not think hold, and I would rather explain than change the code quietly.

On bounding the query to the simulated date: `simulate_backfill` calls every fetcher as `fetcher(source_config, earliest_since, limit)` without `_collection_now`, and it has always done so for all of `BACKFILL_SOURCES`, Crossref included. `_latest_allowed` then falls back to the real now, which is the correct upper bound for a fetch, and the simulated date is applied afterwards in the per-date loop that keeps only items where `since <= (updated_at or published_at) <= simulated_now`. Adding OpenAIRE to that set inherits the existing behaviour rather than introducing a new gap; changing it would be a change to the shared backfill contract and belongs in its own PR.

On the evidence credit: `EVIDENCE_PRIMARY_SOURCES` already lists arXiv, GitHub, Hugging Face, OpenReview, Semantic Scholar and Crossref, and every one of them receives the same 40-point credit for the same reason. OpenAIRE is a primary research-product source of exactly that kind, so listing it follows the existing rule. If the rule itself should be evidence-quality-weighted rather than source-membership-based, that is a worthwhile change but it would apply to all seven sources, not to this connector alone.

ktwu01:
Agent annotation: `330226` / Claude

Pushed `c841a81` to this branch. The fix came out of calling the live Graph v3 endpoint, which the PR description correctly flagged as the one thing that had not been exercised.

## The bug

`fetch_openaire` read `row.get("description")`. v3 returns the abstract as `descriptions`, a **list**, and carries no singular `description` member at all. Sampling 8 live products: 8/8 expose `descriptions`, 0/8 expose `description`. Every record this connector published would have reached the site with an empty summary.

The suite did not catch it because `_openaire_row` invented the singular key and `test_openaire_preserves_upstream_metadata_and_bounds_the_query` asserted against that invention. The failure was silent rather than loud: `score_item` matches the taxonomy over `title + summary` (`pipeline.py:246`), so records would have been categorised on title alone, and the `title-only provenance` structural signal would not fire because `artifact_urls` and `metrics` are populated.

## The fix

Read `descriptions`, take the first entry that survives `clean_card_text`, treat `null` as an absent abstract, and raise on a non-list or on non-string entries.

Live check after the fix, for `"benchmark dataset"`: of 6 products returned, the 2 that carry an abstract now have summaries of 1,940 and 1,695 characters. The other 4 carry `descriptions: null` upstream, so they are correctly empty rather than being missed.

Fixtures now match the real payload shape, plus two regression tests: a multi-entry `descriptions` with `<jats:*>` markup and a blank leading entry, and `descriptions: null`.

## Two documented claims corrected

Both are contradicted by the live API:

- The anonymous budget is 7,199 requests/hour (reported in `x-ratelimit-limit`), not 60.
- An unquoted multi-word value returns 200, not 400. Quoting is still the right call, but for precision rather than acceptance: `LLM benchmark` returns 213 titles quoted against 3,021 unquoted.

## Verification

Full `ci.yml` sequence on a clean worktree: `ruff check`, `ruff format --check`, `normalize-catalog`, `classify`, `build-data-release`, then `pytest -q`. Result: **1417 passed, 0 failed** (an untouched `main` baseline in the same sandbox: 1362 passed).

## Still needs a maintainer

CI has never run on this PR. Every run sits at `action_required` because this is a fork PR, so it needs workflow approval before any check can report.

Two items left for the author, not changed here:

- The merge-order note references `#544`, which is the issue number. The DataCite PR is `#613`, and both touch the same 11 files.
- Expected yield is low. Across 2026-08-01 to 2026-09-13 the six phrases matched 0, 1, 1, 5, 25 and 0 titles respectively, and `lookback_hours` is 48, so most daily runs will collect nothing. That is a value question rather than a defect.

ktwu01 (changes requested):
Thanks for the live API follow-up; the `descriptions` and missing-`results` fixes are sound. I still found a few data-contract blockers:

- Missing counters are published as zero. Please omit absent citation, download, and view metrics while preserving a real upstream zero.
- Non-DOI PIDs and `dateOfCollection` exist only in `raw`, which `RadarItem.to_dict()` removes, so the required identifiers and dates do not reach durable snapshots. Please project them into the public record contract and test serialization.
- Historical backfill reads only one newest-first page per phrase, so longer spans silently lose older OpenAIRE records while reporting healthy. Please paginate backfill or exclude this capped connector from it.
- Date-only rows can pass as not-future during the five minutes before midnight; nested non-string author or organization names are also converted to Python representation text. Please add direct guards and regression cases.

Both clean CI runs pass (1,418 tests on the head; 1,440 with current `main`) and the merge itself is conflict-free. Thank you.

ktwu01:
The live-API follow-up (`c841a81`) is solid. Two blockers and two things to
narrow.

**Blockers**

1. **Non-DOI PIDs and `dateOfCollection` live only in `raw`**, which
   `to_dict()` strips, so they never reach the public record. What I would like
   to understand before you change anything: does anything downstream need
   them? The repo rules do not require exporting every upstream field, so the
   case for projecting them has to be an identity or provenance one. If a
   non-DOI PID is the only stable handle for some OpenAIRE records, say so and
   project them with a serialization test. If not, this can close as
   intentional.
2. **Non-string author or organization names stringify to a Python repr.** Same
   shape as the creator-name issue on #613. Add shape validation and a case.

**Narrowing two things I had queued**

- The near-midnight future check: this uses the repository's deliberate
  five-minute `FUTURE_TIMESTAMP_TOLERANCE` (`sources.py:23`), which is a
  repo-wide policy rather than an OpenAIRE defect. If date-only values need
  different handling than timestamps, that argument needs making on its own
  terms; I am not making it here.
- The single newest-first page per phrase: #613 makes the same choice with the
  same reasoning ("one page per search keeps the request budget fixed"), so
  this cannot fairly be an OpenAIRE-only objection. If one-page backfill is
  misleading it is misleading for both, and that is a conversation for the two
  PRs together.

On the absent-vs-zero counters at `sources.py:1285`: not a blocker on your PR.
`principle.md` does forbid treating unknown counts as zero, and the existing 19
occurrences across ten connectors on `main` are precedent rather than
compliance. But making OpenAIRE alone distinguish absent from zero would leave
the corpus inconsistent. I am filing this as a repo-wide issue covering all the
call sites.

Nits: the merge-order note cites #544, which is the issue; the DataCite PR is
#613. And both of you plus #632 bump the README source count, so the last to
land takes 40/16.

---

ktwu01:
Filed the repo-wide counter issue referenced above: #676. It lists all 19 call sites across the ten connectors on `main`, and notes that this PR's instances follow the existing convention rather than introducing the problem.

Issue #589, opened 2026-09-09
[6 points] Past 30 days: no REAL traction signal of the top ranked bench. not know why... Stars: 0 Forks: 0


ktwu01:
Ref to Claire's repo & Code: https://github.com/Claire1217/benchmark-radar/
Ref to Claire's rank: 
现在的排名主要看三个真实数据：Hugging Face Paper 的点赞、GitHub Stars，还有 Hugging Face 数据集的下载量。这三个数据在当日、30 天和 90 天排名里，占的比例是不一样的。

当日主要看 HF Paper 点赞，因为一个 Benchmark 刚发布的时候，GitHub Stars 和数据下载量一般都还没起来，这时候HF 点赞更能代表它有没有马上引起大家的注意。除此之外我发现当日的数量有时候比较有限，就添加了个大模型评估附加分，llm根据论文内容、Benchmark 的覆盖范围，代码和数据是否完整，预测它发布后 7 天内可能获得的关注。

30 天和 90 天的排名，我们会更看重 GitHub Stars 和数据集下载量，时间拉长HF 点赞就没那么重要了。star和下载量数据更能说明大家是不是还在持续关注，以及有没有真的被社区使用。

attention部分还有一些设计细节 比如先看每个数据在同一批 Benchmark 里面排在什么位置，而不是用绝对值；对于缺失数据的处理等。

But OURS are TO BE IMPROVED.

Date:

Past 30 days

Filters
(0)

Past 30 days
5230 normal · 67 attention
Sort: Date, then new releases, then Recency ↓, then Priority ↓
August 11, 2026 – September 9, 2026 · Data through September 9, 2026 UTC

01
Lattenot/agent_benchmark_test
GitHub
Released 8h ago
Agentic
Organizations: Lattenot
Stars: 0
Forks: 0
No description published at the source.

37
/ 100

Priority score
i
02
CodeCrafterAdi2006/AAROH_Platform123
GitHub
Released 8h ago
Benchmark
Agentic
Organizations: CodeCrafterAdi2006
Stars: 0
Forks: 0
AAROH — Hybrid Classical-Quantum Clinical Benchmarking & Decision Support Platform. Orchestrates XGBoost & 4-Qubit VQC via deterministic agentic SSE streaming, TreeSHAP explainability…

45
/ 100

Priority score
i
03
ayush1192/Startup_Feasibility_Analysis
GitHub
Released 8h ago
Benchmark
Dataset
Organizations: ayush1192
[...]

Issue #530, opened 2026-09-03
[12 points] Show the latest leaderboard by default (past 30 days)


ktwu01:
## Outcome

Opening **Leaderboard** should immediately answer one question: **which newly released benchmarks are receiving the most public attention now?**

The default view should be **Latest releases · 30 days**, with the Top 10 visible before methodology or pipeline detail. The existing model-card adoption ranking and reported-score workbench remain available as separate evidence views; their cumulative counts must not be mixed into the recent-release ranking.

中文目标：用户打开 Leaderboard，先看到最近 30 天发布、目前最受关注的 benchmark。现有的 model-card 采用榜和分数历史继续保留，但要明确分开。

@junjiezhou1122 could you review or lead the data-contract and ranking implementation? The first pass should establish reproducible data and tests before changing the default UI.

## Why this is needed

The current leaderboard answers the cumulative question introduced in #83: which benchmarks appear in the tracked model cards. That is useful, but it changes slowly and does not answer what is new now.

#479 asks for an August Top 10 with stars, upvotes, downloads, source coverage, and an explanation of why each benchmark matters. This issue turns that one-month study into a maintained product surface.

Claire's implementation is a useful reference:

- [Methodology](https://github.com/Claire1217/benchmark-radar/blob/main/docs/METHODOLOGY.md)
- [Dated metric collection and window ranking](https://github.com/Claire1217/benchmark-radar/blob/main/pipeline/enrich_metrics.py)
- [Latest / 30 days / 90 days UI](https://github.com/Claire1217/benchmark-radar/blob/main/web/app.js)

We should reuse the dated observations, explicit source status, and release-window separation. We should not copy its LLM forecast, neutral-value imputation for missing signals, or describe cumulative counters as 30-day growth.

## Product contract

Leaderboard has three clearly labelled modes:

- **Latest releases** — default mode, default window **30 days**.
- **Model-card adoption** — the existing #83 cumulative ranking over tracked source documents.
- **Reported scores** — the existing protocol-aware score and per-benchmark workbench.

Latest releases offers `7d`, `30d`, and `90d`. `/leaderboard/` resolves to `Latest releases · 30d`; explicit permalinks retain the chosen mode and window.

The 30-day window is exactly:

```text
window_start = generated_at_utc - 30 * 24 hours
window_start <= release_timestamp <= generated_at_utc
```

[...]

PR #617, opened 2026-09-12
feat(site): open the leaderboard on the latest releases

lazizbekravshanov:
Part of #530: the page half. Together with the collector PR for #589 it completes the issue's acceptance list; the two are independent and merge in either order.

One commit on top of upstream `main` at 5d9e378.

## What changes for a reader

- **`/leaderboard/` opens on "Latest releases · 30 days".** Recently released benchmarks ranked by the attention they are receiving now, from the `latest_releases_leaderboard` payload the pipeline already publishes. 7- and 90-day windows are one click away.
- **Model-card adoption and reported scores stay separate modes on the same address.** The adoption view keeps every element, id, seed and handler it has today, shown as a block when the reader picks it. Reported scores links to the existing `/saturation/` workbench.
- **Every rank shows its inputs.** Each row prints rank, name, release date, a bar scaled against the top score, the score and its confidence. Opening a row shows each signal with its value and status (fresh, stale since the day it was last read, unavailable, or not observed), its weight and a link to the resource it was read from, plus coverage and confidence. A release the engine could not rank stays in the list, marked as limited signals, rather than vanishing.
- **The (i) states the method.** Method version, the weights read from the published components (not restated in the browser), the log1p normalization, that downloads are a rolling 30-day figure and never cumulative, that stars come from the benchmark's own repository and never a parent framework, and the window bounds in UTC.
- **An empty window is a real state.** The page says no release in the window has a measurable signal yet and offers the wider windows; past the widest one it offers the adoption view. A window the corpus fails to load, or does not carry, says so instead of sitting on a loading line, and the refresh control re-requests the corpus for it.
- **Switching modes is one Back away.** A mode is a different ranking, so choosing one pushes a history entry; picking a window refines the entry in place. Opening a row's signals survives a redraw (the catalog index landing, a refresh, Back).
- **Chinese interface.** Every visible string has a translation.

## Acceptance criteria this covers

| Criterion | Where |
| --- | --- |
| `/leaderboard/` defaults to Latest releases · 30d | mode strip, `state.lmode`, `latestReleasesDefaultWindow` |
| Each rank exposes inputs, components, coverage, confidence, method version, source links | `latestReleaseRow`, `latestReleasesMethodNote` |
| Missing and stale metrics stay distinguishable | `latestSignalText` |
| Downloads never presented as cumulative or delta | signal label "last 30 days" and the method note |
| Existing views preserved with current trust labels | adoption block unchanged; `/saturation/` unchanged |
| Empty states suggest broader windows | `latestReleasesEmptyState` |

The cohort, dedupe, keyword and parent-framework rules are the engine's (`release_leaderboard.py`) and the collector's (the collector PR); this PR prints what they publish and computes nothing.

## Addresses

A permalink written before the page had modes carries the adoption view's own filters (`lscore`, `lq`, `ldomain`, `lorg`, `lera`, `lheight`). Any of them selects the adoption mode on read, and the mode writes them back in exactly the shape it always did, so `test_clean_route_model_migrates_legacy_urls_and_preserves_utility_backgrounds` holds unchanged. The latest mode writes only a non-default `lwindow`, so its address can never flip modes on reload. `?lmode=latest|adoption` is accepted as an explicit override.

The cutoff is shared with Saturation, whose every address carries it, so on a Saturation address `lscore` alone is not a mode choice: a reader who goes Latest releases → Saturation → Back → Forward → Leaderboard lands on Latest releases, not on a mode they never picked. The mode is resolved after the legacy benchmark permalink redirect for the same reason.

## Wiring

- `site/index.html`: the mode strip, the latest-releases section, and the adoption block wrapper (`#leaderboard-adoption`, hidden by default; its seeded rows stay in the document for crawlers).
- `site/assets/app.js`: `state.lmode` / `state.lwindow`; `readUrl` and `writeUrl`; `leaderboardModeFromParams`, `latestReleasesWindowKey`, `latestSignalText`, `latestReleasesEmptyState`, `latestReleasesMethodNote`, `latestReleaseRow`, `renderLatestReleases`, `syncLeaderboardMode`; one delegated click handler for the mode and window controls (the empty state renders its own buttons); zh strings.
- `site/assets/styles.css`: the mode and window pills, the row disclosure and its signal grid, phone layout.
- `src/benchmark_radar/app_seeds.py`: `_latest_releases_seed`, so the static `/leaderboard/` page carries what the renderer draws for the default window: the rows with their disclosure bodies (signals, statuses, weights, source links, coverage and confidence), the method note, and, for an empty window, the empty state with its way out. A test runs the real renderer on a shared fixture and compares the markup byte for byte with the seed.
- `site/llms.txt` and `VIEW_SEO.leaderboard`: the page's title, description and llms.txt entry describe the new default.
- `tests/test_site.py`: the route-model test loads the mode and window helpers `readUrl` and `writeUrl` now call.
- The bootstrap payload already carries only the default window (`snapshots.py`); a leaderboard address naming another one is a full-data route in `stateNeedsFullData`, so boot, Back and the refresh control fetch the corpus for it the way they do for a historical Today URL, and the renderer fetches it on a window click.

## Tests

`tests/test_latest_releases_view.py` (eleven tests), `tests/latest_releases_harness.mjs` and `tests/latest_releases_render_harness.mjs`, each named for the failure it protects against: the page opens on the latest ranking with the adoption block hidden and intact; mode resolution for bare, legacy, explicit and bogus addresses, and for a Saturation address with and without the adoption filters; window fallback; signal text for fresh, stale (dated and undated), unavailable, zero and null; the empty-state suggestion for a loaded-empty, unloaded and widest window; weights read from components; the method note; the URL round trip through the real `readUrl` and `writeUrl`; adoption filters written only in adoption mode, a mode switch pushed and a window switch replaced; every visible string translated and no key entered twice; the static seed (order, escaping, limited entries kept, disclosure bodies, method note, empty state, default window named, nothing seeded for nothing); the seed equal to what the real renderer draws on a shared fixture (a ranked row with fresh, stale and unavailable signals, a limited row, an unsafe source URL, an empty window); the renderer keeping open rows across a redraw, treating a missing window as a full-data route, fetching it, reporting a failed fetch and a window the corpus does not carry, and drawing the Chinese interface; the page metadata describing the new default; the written page carrying both rankings; bar widths scaled against the top score.

Twenty-two behaviours were checked by reverting each one and watching its test go red, among them: the default mode, adoption filters leaking into the latest address, Saturation's cutoff choosing a mode, the mode resolved before the legacy redirect, a missing window not counted as a full-data route, a failed fetch left on the loading line, a loaded corpus without the window left on it, open rows collapsing, a mode switch replacing history, the seed dropping the confidence pill, the disclosure body, a weight, or offering a window known to be empty, a release date formatted outside UTC, a duplicated translation key, and the method note printed for a window with no bounds.

## Verification

Clean-worktree run of the CI sequence (`git worktree add --detach`, submodule initialized): `ruff check .`, `ruff format --check .`, `normalize-catalog`, `classify` and `build-data-release` pass; `pytest -q` reports 1344 passed, 1 failed.

Driven in Chromium against a built payload with four synthetic entries (three ranked, one limited): the page opens on the 30-day ranking with the adoption block hidden and the unranked row announced as "unranked" to assistive technology; a row opens to its signals, with the stale upvote dated and the gone dataset marked unavailable; the 7-day window fetches the full corpus and lands on the empty state whose "Try 30 days" button restores the ranking; the 90-day window shows its five rows, and booting directly on `?lwindow=90d` loads them through the boot path; choosing the adoption mode pushes one history entry (pressing it again pushes none) and Back returns to Latest releases; an opened row stays open across the refresh control; `/saturation/?lscore=60` then the Leaderboard tab opens Latest releases, while `/saturation/?lscore=60&lq=agent` opens adoption with the search filled; "Reported scores" switches to Saturation without a page load; with `radar.json` blocked, the 90-day window reports that it could not be loaded and the refresh control brings its rows back, and booting on the 90-day address shows the error banner rather than a loading line; the Chinese interface translates the heading, modes, columns, note and signal text; at 400px the body has no horizontal scroll. No page errors.

## Not in this PR

The collector that fills the payload with real counters (the collector PR). Until it merges the live page shows the empty state, which is the honest reading of an empty payload.

### DIFF
--- a/site/assets/app.js
+++ b/site/assets/app.js
@@ -1116,6 +1116,49 @@ const I18N = {
     release: "发布",
     "release date unrecorded": "未记录发布日期",
     released: "发布于",
+    // Latest releases (issue #530)
+    "Leaderboard views": "排行榜视图",
+    "Latest releases": "最新发布",
+    "Model-card adoption": "模型卡采用情况",
+    "Reported scores": "报告分数",
+    "latest.deck":
+      "近期发布的benchmark，按当前获得的关注度排序：GitHub star、Hugging Face 论文点赞以及最近 30 天的 Hugging Face 数据集下载量。",
+    "Release window": "发布时间窗口",
+    "{days} days": "{days} 天",
+    "7 days": "7 天",
+    "30 days": "30 天",
+    "90 days": "90 天",
+    "Attention score": "关注度得分",
+    "GitHub stars": "GitHub star 数",
+    "Hugging Face paper upvotes": "Hugging Face 论文点赞数",
+    "Hugging Face dataset downloads, last 30 days": "Hugging Face 数据集下载量（最近 30 天）",
+    unavailable: "资源已不可用",
+    "not observed": "未观测到",
+    "{value} · stale since {date}": "{value} · 自 {date} 起未更新",
+    "{value} · stale": "{value} · 未更新",
+    "{value} · unverified since {date}": "{value} · 自 {date} 起未经核验",
+    "{value} · unverified": "{value} · 未经核验",
+    "source ↗": "来源 ↗",
+    "Coverage {coverage}": "覆盖率 {coverage}",
+    unranked: "未排名",
+    high: "高",
+    medium: "中",
+    low: "低",
+    "limited signals: not enough fresh durable signal to rank":
+      "信号有限：缺少足够新鲜的持久信号，未纳入排名",
+    "no score": "无得分",
+    "No benchmark released in the last {days} days has a measurable attention signal yet.":
+      "最近 {days} 天内发布的benchmark尚无可测量的关注度信号。",
+    "Try {window}": "试试 {window}",
+    "See model-card adoption instead": "改看模型卡采用情况",
+    "Loading this window…": "正在加载该时间窗口…",
+    "This window could not be loaded. Refresh to try again.": "无法加载该时间窗口，请刷新重试。",
+    "The {window} window is not in this build.": "此版本未提供 {window} 的时间窗口。",
+    "The latest releases ranking is not available in this build.": "此版本未提供最新发布排行。",
+    "{ranked} of {total} releases in this window are ranked; the rest are listed with limited signals.":
+      "该时间窗口内 {total} 个发布中有 {ranked} 个进入排名；其余因信号有限仅列出。",
+    "Ranking {method}: {signals}, each normalized as log1p(value) / log1p(window maximum) and summed to a 0 to 100 score. A release is ranked only when enough of its weight comes from fresh, durable signals; the rest are listed with limited signals. Dataset downloads are a rolling 30-day figure, never a cumulative total. Stars come from the benchmark's own repository, never a parent framework. Window {start} to {end}, UTC.":
+      "排名方法 {method}：{signals}，各信号按 log1p(值) / log1p(窗口最大值) 归一化后加权求和为 0 到 100 的得分。只有当足够权重来自新鲜且持久的信号时才纳入排名，其余因信号有限仅列出。数据集下载量为滚动 30 天数字，绝非累计总量。star 数来自该benchmark自己的仓库，绝非上级框架。时间窗口 {start} 至 {end}（UTC）。",
     "scale. Every number below is read from the same definition the pipeline applies.":
       "的标尺。下面每个数字都按流程应用的同一套定义读取。",
     to: "到",
@@ -1221,6 +1264,11 @@ const state = {
   benchmarkQuery: "",
   leaderboardShowAll: false,
   leaderboardTopExpanded: false,
+  // Issue #530: the leaderboard opens on the latest releases; the cumulative
+  // adoption view is one click, or one legacy permalink, away. An empty
+  // window means "the payload's default".
+  lmode: "latest",
+  lwindow: "",
   todayResultsKey: "",
   todayRenderedDate: "",
   todayPage: 1,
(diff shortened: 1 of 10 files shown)


ktwu01 (changes requested):
Thanks for the careful implementation and seed-parity work. The combined tree passes all local CI checks (1,395 tests), but several reader-facing #530 contracts are still missing:\n\n- Render each component’s published normalized value, including an explicit unknown state, and mirror it in the static seed.\n- Put each benchmark’s one-line purpose in the visible row summary; it is currently only inside the collapsed disclosure.\n- Include data-through and signal coverage in the compact scope note for the active window.\n- On a direct non-default-window URL, render or clear that requested window when the full-data fetch fails; otherwise the 30-day static seed can remain under a 90-day URL.\n- Add search for the current latest-release cohort rather than leaving search only inside the hidden adoption view.\n\nCould you address these and extend the renderer and seed tests before merge?

ktwu01:
Thanks for the seed-parity work and the revert-checked behaviors. Five items
from #530 still read as outstanding on this head:

1. Render each component's published `normalized` value, including the explicit
   unknown state, and mirror it in the static seed. The fixtures carry
   `normalized` but the rows print value and weight only.
2. Put each benchmark's one-line `purpose` in the visible row summary; it is
   currently disclosure-only.
3. Include data-through and signal coverage in the scope note.
4. On a direct non-default-window URL whose full-data fetch fails, the 30-day
   seed can stay rendered under a 90-day URL. Make the failure visible rather
   than silently serving the wrong window.
5. Add search over the latest-release cohort.

(Items 1 through 3 and 5 are carried from the previous pass against an
unchanged head and I did not re-exercise the UI. Item 4 I traced in the
bootstrap path this round.)

---

PR #601, opened 2026-09-11
Cite the arXiv paper across all citation surfaces

ktwu01:
## Summary
- The technical report is now on arXiv (arXiv:2609.11115, cs.AI, cross-listed cs.IR). Every citation surface previously pointed at the deliberately frozen Zenodo v0.9.0 deposit with only 2 of the paper's 7 authors listed.
- Updates `citation.py`, `app_seeds.py`, the client-side `app.js` mirror, `CITATION.cff`, both READMEs, and the homepage's Highwire citation meta tags to the arXiv identifiers (`10.48550/arXiv.2609.11115`) and the full 7-author byline (Wu, Zhou, Shang, Wang, Han, Wang, Xu).
- The frozen Zenodo v0.9.0 deposit itself (`AGENTS.md`, `zenodo-metadata.json`, the archival PDF, and its pinning test) is untouched — it remains a legitimate historical record, just no longer the preferred citation.
- APA reference-list entries list all authors up to 20 (no "et al." truncation there — that only applies to in-text citations), so all 7 authors appear in the APA and BibTeX forms.

Closes #566, #464, #541.

## Test plan
- [x] `ruff check .` / `ruff format --check .`
- [x] `pytest tests/test_readme.py tests/test_site.py tests/test_query_surfaces.py tests/test_app_pages.py tests/test_technical_report.py -q` (216 passed)
- [x] Opened `/cite/` and `/` locally in a browser to confirm the citation block renders the new APA/BibTeX content correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #618, opened 2026-09-12
Cite the paper inside the query payload and llms.txt

junjiezhou1122:
## Summary

#601 moved every citation surface to the arXiv report, but two agent-facing paths still missed it.

- **`data.citation` now travels in every payload command** (`search`, `show`, `recent`, `status`). It rides the single `_provenance()` path rather than a per-command patch, so a new command cannot forget it. It carries APA, BibTeX, and the `\cite{}` key.
- **`site/llms.txt` names the paper** and its arXiv ID next to the home page, and the citation entry now describes what the `/cite/` page actually offers.
- **`citation.py` owns the BibTeX entry**, and a test ties the five hand-copied surfaces (both READMEs, `CITATION.md`, `app_seeds.py`, `app.js`) to it.

### Why the payload, not just the footer

The reminder rides stderr so `--json` stdout stays byte-identical to the HTTP contract (#483). An agent that reads stdout alone — a `jq` pipe, a framework that captures one stream — therefore never receives the paper. Now it does, in a form it can put straight into a related-work table instead of looking the paper up again.

`citation` sits beside `results` under `data`, so it cannot displace a candidate. No interface-specific ranking, no new command, no change to filters or the result array.

### Why the BibTeX test

#601 shipped the arXiv entry to five surfaces in one commit and left nothing comparing them, so the next edit can update one and silently stale the rest. `citation.py` now holds the entry instead of a sixth copy, and each field is compared across all five.

### Files

`citation.py`, `query.py`, `site/llms.txt`, `tests/test_query_surfaces.py`, `tests/test_readme.py`, `tests/test_site_seo.py`. Two existing assertions pinning `data == {"source": "local"}` exactly were updated to include `citation`.

## Test plan

- [x] `ruff check .` / `ruff format --check .`
- [x] `benchmark-radar normalize-external` → `classify` → `build-data-release`, in order
- [x] `pytest -q` → **1337 passed, 0 failed**
- [x] Ran `search` / `show` / `recent` / `status` in both human and `--json` modes; confirmed `serve` returns it on `/api/v1/search` and `/healthz`


toby-bridges (approved):
Reviewed commit 28b9e62d4a7791165c78d710c22038d1f014ec57. No blocking findings.

Compared search, show, recent and status against main using the same rebuilt corpus: each response adds data.citation, with all other fields unchanged. The paper title, author list and arXiv identifier match the published record.

Validation in a fresh detached worktree with the pinned paper submodule initialized: ruff check ., ruff format --check ., benchmark-radar normalize-catalog, benchmark-radar classify, benchmark-radar build-data-release, then pytest -q. All six steps passed; 1,337 tests passed, 0 failed. Local environment: macOS, Python 3.12.13, Node 24.19.0; GitHub CI uses Node 22 and is also green.

PR #619, opened 2026-09-12
Add a model card SOP and split AGENTS.md into reference guides

ktwu01:
## Summary

`AGENTS.md` is read in full on every task, so reference material that applies to a minority of changes was competing for attention with rules that apply to all of them. This splits the reference material out and adds the procedure that was missing.

**New: [`docs/sop-add-model-cards.md`](docs/sop-add-model-cards.md).** Adding a model card spans two files and a set of loader checks that no single place stated. A contributor had to reconstruct it from `CONTRIBUTING.md`, the header of `data/model_cards.yml`, and the header of `data/benchmark_scores.yml`, and in practice the scores got missed. The SOP walks the whole change: register the document, add any missing benchmark, add the score rows, rebuild. Field tables and enums come from the loaders (`model_cards.py`, `benchmark_scores.py`) rather than from current data, so required versus optional is accurate. It closes with what the loader rejects and, separately, what it cannot catch.

**New: [`docs/pipeline-and-data-map.md`](docs/pipeline-and-data-map.md) and [`docs/query-surfaces.md`](docs/query-surfaces.md).** Moved out of `AGENTS.md`. No rule text changed. Headings shift one level, and the technical report section now names `AGENTS.md` for the CI sequence instead of saying "below".

**`AGENTS.md` loses 176 lines of reference material and gains a 22-line routing table**, leaving it at 130 lines. Each row of that table says when a guide is required reading and which constraint holds whether or not you open it, so moving the detail out does not quietly drop the rule.

**Two contributor rules stay inline**, because they apply to every change: a model card PR must carry its scores, and an agent opening a PR self-identifies with the marker `330226` plus its model series.

**`CONTRIBUTING.md`** now points at the SOP. Its "no scores in the adoption registry" rule reads as "no scores wanted" without naming the archive that does take them.

## Test plan

Documentation only. No Python and no data files are touched.

- [x] `ruff check .` passes
- [x] `ruff format --check .` passes, 168 files already formatted
- [x] Every line removed from `AGENTS.md` is verified present in a new doc, or accounted for as an intentional rewrite: the inline subsection became the SOP, one cross-reference was retargeted, six headings shifted level
- [x] All five relative documentation links resolve
- [x] Markdown is not linted in `ci.yml`, and no markdownlint or prettier config exists
- [x] The generator sequence and `pytest` cannot be affected by a Markdown-only diff, so they were not re-run

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,6 +7,19 @@ reduction to a few dozen. Missing measurements must not remove corpus records.
 Benchmark Frontier and its linked score ranking explicitly exclude records
 without numeric reported scores, as specified in `principle.md`.
 
+## Reference guides
+
+This file holds the rules that apply to every change. Detailed procedure lives
+in the guides below. Each line states when the guide is required reading, and
+the constraint that holds whether or not you open it.
+
+| Guide | Read it before | Constraint that always holds |
+| --- | --- | --- |
+| [`docs/pipeline-and-data-map.md`](docs/pipeline-and-data-map.md) | Touching a source input, a generator, a generated artifact, or the technical report | Never reconstruct the system from a report, the deployed site, or leftover generated files. Never patch a derived file to fix a source-data problem. |
+| [`docs/sop-add-model-cards.md`](docs/sop-add-model-cards.md) | Adding a model card, a benchmark, or a score | `data/model_cards.yml` and `data/benchmark_scores.yml` move together. Every value is read out of the cited document, never from memory. |
+| [`docs/query-surfaces.md`](docs/query-surfaces.md) | Changing search, detail lookup, the CLI or HTTP query surface, or the consumer Skill | `QueryService` is the single source of truth. No interface-specific ranking, and no silent network fallback. |
+| [`principle.md`](principle.md) | Changing any benchmark-facing surface | Start from the full corpus across all sources. |
+
 ## Glob rule: showcase and UI communication
 
 Applies to `README*`, `docs/**`, `.github/ISSUE_TEMPLATE/**`, `site/**`, and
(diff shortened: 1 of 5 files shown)


ktwu01:
Agent-authored PR: 330226

Model series: Claude (Opus 5, via Claude Code)

Annotating per the convention this PR adds to `AGENTS.md`.

ktwu01 (commented):
okay

ktwu01:
@codex review

Addressed the five findings from the previous review in 7b3fabd. Documentation only; no Python and no data files.

- **`reported_by` provenance.** Confirmed the mechanism you described: `benchmark_scores.py` defaults the field to `None`, and `catalog_reports.py` then emits `reported_by: "self_reported"` with `measured_by` falling back to the scored model's own `organization`. The SOP now states the field is optional to the loader and mandatory in fact, and names the consequence of omitting it.
- **Absolute score requirement.** Scoped to numbers that can actually be read, in both `AGENTS.md` and the SOP intro. A card whose results are only qualitative, or whose table is an unreadable image, is now explicitly a finished change with no score rows, which is what the SOP's own recording rule already said.
- **`caveat` guidance.** The residual was in "what the loader cannot catch", which asked for what the benchmark *measures* and so pointed away from the comparison warning described earlier in the file. Restated as a warning about what would mislead someone comparing two reported numbers.
- **`document_type`.** Documented that it must be set whenever the document is not literally a model card, since the loader accepts its absence and `adoption_rank` defaults it. For the record, current data is clean: all 37 registered cards set the field, so this is a guide gap rather than a data problem.
- **Metric and unit.** Result rows carry neither field and a benchmark holds exactly one metric, so "one row per reported number" is now "one row per number reported in the declared metric", with a note that a second metric for the same `benchmark_id` is rejected and those numbers should be left out rather than mislabelled.

`ruff check .` and `ruff format --check .` pass. All relative documentation links resolve.

ktwu01 (commented):
okay

PR #625, opened 2026-09-13
[不要自动merge] fix(site): halve the leaderboard title and anchor its info toggle

ktwu01:
Issue #313. The 48px uppercase heading was the largest object on the page and compressed the ranking it names; worse, the wrapped title filled its whole max-width box, so the blue (i) rendered hundreds of pixels away with no anchor. Half scale fits the title on one line, letting flexbox place the toggle immediately after the last glyph, and flex-wrap stacks them on narrow screens instead of drifting.

### DIFF
--- a/site/assets/styles.css
+++ b/site/assets/styles.css
@@ -2311,9 +2311,22 @@ td a {
 
 .leaderboard-top-heading {
   display: flex;
+  flex-wrap: wrap;
   align-items: center;
   gap: 0.4rem;
   margin-bottom: 0.35rem;
+}
+
+/* Half the global h1 scale (issue #313). The old 48px uppercase line was the
+   largest object on the page and compressed the ranking it names; and because
+   the wrapped title filled its whole max-width box, the (i) beside it landed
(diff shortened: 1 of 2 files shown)


ktwu01 (commented):
Thanks for preserving this fix. Current  already contains the #313 behavior: the heading row wraps, and both leaderboard headings use the reduced  scale with the info control anchored in the same flex row.\n\nMerging this older branch now produces conflicts in the stylesheet and leaderboard test file. Since the requested behavior is already live on , I recommend closing this PR as superseded rather than resolving and merging it.

ktwu01 (commented):
Formatting correction to my comment above:

Current `main` already contains the #313 behavior: the heading row wraps, and both leaderboard headings use the reduced `clamp(1.25rem, 2vw, 1.5rem)` scale with the info control anchored in the same flex row. Merging this older branch now produces conflicts in the stylesheet and leaderboard test file.

Since the requested behavior is already live on `main`, I recommend closing this PR as superseded rather than resolving and merging it.

Issue #380, opened 2026-08-26
[12 points] 参考 xbsleepy's methodology


ktwu01:
https://xbsleepy.github.io/daily-agent-benchmarks/

Need at least GPT 5.6 Sol to do this
- [ ] use GitHub action to routinely get their data to be one source of ours
- [ ] We can classify the benchmarks based on their classification
- [ ] Add a UI of classification, just like theirs
- [ ] Make sure to follow our own pipeline, don't reinvent wheel!

https://www.xiaohongshu.com/discovery/item/6a8db6b20000000003028f24?app_platform=ios&app_version=9.41&share_from_user_hidden=true&xsec_source=app_share&type=normal&xsec_token=CBwfPuiOOkPMJayxPGqR9sYhJB76Ym7Pa2_ha1NEEGFNA=&author_share=1&xhsshare=WeixinSession&shareRedId=ODdHNkk1RTs2NzUyOTgwNjczOTc8OTZA&apptime=1787720407&share_id=92124294fbc140bbacc2780cceacf9e1

PR #632, opened 2026-09-13
Add XBsleepy daily-agent-benchmarks digest connector (issue #380, part 1/3)

Jeffyaoliang:
Part one of #380 — the data-source half. The theme-classification UI (part three) is deliberately left for a follow-up so this PR stays one logical change.

## What

- `fetch_xbsleepy` connector in `src/benchmark_radar/sources.py`: ingests the digest's published cumulative index JSON (`xbsleepy.github.io/daily-agent-benchmarks/data/index.json`), filters locally by `since` (an unchanged index is a healthy empty fetch), and maps papers onto `RadarItem` with authors, abs/pdf/html artifact URLs, citation counts, and the digest's theme tags.
- Theme tags ride through in `categories` prefixed `xbsleepy:` (e.g. `xbsleepy:planning`) so downstream grouping can use the digest's own taxonomy without colliding with keyword-derived categories. The raw digest row (incl. `field` and match hints) is preserved in `RadarItem.raw`.
- Registered in `SOURCE_FETCHERS`, `_PARSER_VERSION_METHODS` (`xbsleepy-index → Index JSON`), `SOURCE_DEFAULT_METHODS`, and `config.yml` as an **optional, allow-empty** source.
- No new GitHub Action: the existing daily run collects every enabled source in `config.yml`, so the digest is swept up automatically — per the issue's "follow our own pipeline, don't reinvent wheel".

## Why

The digest is a high-precision, already-curated feed of arXiv papers that *introduce* agent benchmarks (two-stage classifier in `xbsleepy/daily-agent-benchmarks` `scripts/classify.py`), which complements the radar's own keyword-based arXiv connector and gives the theme grouping in #380 a ground-truth source.

## Verification

- `ruff check .` / `ruff format --check .` clean.
- `pytest tests/test_sources.py`: 83 passed, including 5 new tests covering mapping + theme pass-through, date-only `announced_date` parsing, staleness filtering, missing-`papers` rejection, and future-timestamp rejection via the connector-wide `_collection_now` guard.
- Live smoke against the real index: 10 items mapped for `since = 2026-09-08` with correct source ids, artifact URLs, and prefixed theme categories.
- Full `pytest -q` on this fresh checkout shows the same pre-existing failures with and without this change (stash-verified): they come from gitignored generated artifacts (`site/data/benchmark-index.json` etc.) missing until `normalize-catalog` runs, not from this PR.

## Follow-up

The theme-grouping browse view over `categories` (part three of #380). Happy to take direction on where it belongs — a new site view vs. a section on the Today page.


ktwu01 (changes requested):
Thanks for adding the XBsleepy connector. I found a few blockers on the current  integration:\n\n- The freshness cutoff parses  but does not use it when  or  exists, so a paper can be missed on its first digest announcement. Please include the announcement timestamp in activity and future-date handling while keeping the paper publication date as provenance.\n- Please add a reader-facing  entry for .\n- Please update both READMEs from 37/13 to 38/14 sources/connectors.\n- The missing- test currently fails earlier on a missing ; please make it exercise the payload validation it names.\n\nOn a clean combined tree, Ruff and all three generators pass and the catalog rebuilds to 1,284 records, but pytest ends with 1,387 passed and the two source-name/count tests failing. Could you address these before merge?

ktwu01 (commented):

- The freshness cutoff parses `announced_date` but does not use it when `published` or `updated` exists, so a paper can be missed on its first digest announcement. Please include the announcement timestamp in activity and future-date handling while keeping the paper publication date as provenance.
- Please add a reader-facing `SOURCE_DISPLAY_NAMES` entry for `xbsleepy`.
- Please update both READMEs from 37/13 to 38/14 sources/connectors.
- The missing-`papers` test currently fails earlier on a missing `index_url`; please make it exercise the payload validation it names.

On a clean combined tree, Ruff and all three generators pass and the catalog rebuilds to 1,284 records, but pytest ends with 1,387 passed and the two source-name/count tests failing. Thank you.

ktwu01:
Nice strict connector: the payload validation, the future-date rejection and
the `xbsleepy:` tag prefix all read well. Two correctness blockers, then two
rebase chores.

**Blockers**

1. **`announced_date` is parsed but cannot act.** `sources.py:1573-1580` sets
   `published_at = published or announced`, so the announcement date is only a
   fallback. A paper announced today whose `published` is older and which has
   no `updated` gets `activity_at` in the past and is dropped by the
   `activity_at < since` test. I reproduced the drop against this head.
   `_reject_future` is likewise passed `published_at` and `updated` but never
   `announced`. Include the announcement in both the activity and the
   future-date paths.
   `test_xbsleepy_accepts_date_only_announced_dates` covers the parsing but not
   this case.
2. **`test_xbsleepy_requires_papers_array` does not test what it names.**
   `tests/test_sources.py:2040-2047` passes `{}` as config, so the function
   raises on the missing `index_url` at `sources.py:1561` and never reaches
   `_payload_rows`. The test's own inline comment says as much. Give it a valid
   `index_url` and a payload with no `papers` key.

**Acceptance criterion worth adding.** The `xbsleepy:*` theme tags this
connector attaches do not survive `score_item()`, which overwrites
`item.categories` at `pipeline.py:262`. So the tags are promised here and
discarded downstream. Worth an end-to-end test through the pipeline as part of
this PR, since a connector-level test passes either way. One caution: the fix
is to preserve the tags as tags, not to feed them into relevance scoring.

**On rebase**

- Add `SOURCE_DISPLAY_NAMES` for `xbsleepy` in `site/assets/app.js:47` here
  rather than in the UI PR; until then the source renders as the raw key.
- Update `README.md:160` and `README.zh-CN.md:136` from 37/13 to 38/14. #613
  and #614 also each add a connector, so whoever lands later takes 39/15 and
  40/16.

---

PR #638, opened 2026-09-13
Add theme browse page generated from snapshot categories (issue #380, part 3/3)

Jeffyaoliang:
Part three of #380, stacked on #632 (review/merge that first — this branch is built on it, and the xbsleepy themes it surfaces arrive through that connector).

## What

- **New generator** `src/benchmark_radar/themes.py`: writes `site/themes/` — a browse page grouping discovery records by the categories their source attached. It renders whatever categories arrive instead of hardcoding a taxonomy: the XBsleepy digest's capability themes come in as `xbsleepy:`-prefixed categories from #632, and any source can join by prefixing its own.
- Wired into the Pages build next to `write_blog`, from the same validated snapshots in the same run, so a group can never name a record the published corpus has moved past.
- Reuses the extracted site chrome with the contract running both ways: the dashboard footer must link `/themes/` or the build refuses, and the page marks that footer link active.
- `SOURCE_DISPLAY_NAMES` gains `xbsleepy` so zero-fetch days name the source in the source-mix ledger (issue #260's rule: zeros carry the signal).
- Sitemap entry rides the same build.

## Page design

Receiver-first per the showcase rule: one hero stating what this shows and why it matters (bilingual lede, EN primary + zh), theme chips with counts, per-theme record cards (title link, source · date, clipped summary), a client-side filter, and an honest empty state for checkouts without tagged records yet.

## Verification

- `pytest tests/test_themes.py tests/test_sources.py tests/test_blog.py tests/test_snapshots.py tests/test_site.py`: 346 passed — covers grouping/dedup/uncategorized skips, page rendering, empty state, the chrome contract in both directions, the sitemap URL, and the footer link staying in the dashboard.
- `ruff check .` clean.
- Live smoke against the committed snapshots: the page builds with **8,105 records across 5 category groups**, sitemap lastmod 2026-09-13.
- Full `pytest -q` failure set on this fresh checkout is identical to the pre-existing baseline (gitignored generated artifacts missing until `normalize-catalog`), unchanged by this PR.

## Notes for review

- The page groups by whatever `categories` arrive. Today that means the radar's own taxonomy labels on historic snapshots plus `xbsleepy:*` once #632 lands and a daily run ingests the digest. If you'd rather scope the page to `xbsleepy:*` only, or fold it into the Today view instead of a standalone page, both are small follow-up moves.
- Draft until #632 merges so the two land in order.


ktwu01 (commented):
Thanks for keeping this as a draft while #632 is pending. A few end-to-end issues need correction before it is ready for review.

The connector stores `xbsleepy:*` tags in `RadarItem.categories`, but `score_item()` replaces that list with the local taxonomy, so none of the source themes reach snapshots. The theme chips also compare `theme-<slug>` with `<slug>`, which hides every section when clicked.

Please preserve source-native tags through the pipeline with an end-to-end test, use `(source, source_id)` for deduplication, and build the benchmark-facing browse surface from the unified catalog in line with `principle.md`, keeping uncategorized records accounted for. The exact merge also needs `ruff format` on `themes.py`; targeted tests otherwise pass 300 with 1 skipped. Thank you.

ktwu01:
Thanks for keeping this stacked behind #632. Three end-to-end issues while it
is still a draft:

1. `score_item()` assigns `item.categories = categories` (`pipeline.py:262`),
   replacing whatever the connector supplied. The `xbsleepy:*` theme tags #632
   attaches are discarded before a snapshot is ever written. The page does not
   end up empty, which is what makes this subtle: the pipeline's own generated
   categories survive, so the page groups by those instead of the source themes
   it advertises. Preserve source-native tags through the pipeline and cover it
   with an end-to-end test rather than a connector test. I have raised the same
   point as an acceptance criterion on #632, since that is where the tags are
   promised.
2. Chip pinning compares the two halves of the wrong pair: buttons carry
   `data-target="theme-{slug}"` and sections carry `data-slug="{slug}"`, so the
   lookup never matches and clicking a chip hides every section.
3. Dedup keys drop source identity. Key on `(source, source_id)` and build from
   the unified catalog so uncategorized records stay accounted for under
   `principle.md` rather than leaving the page.

Also needs `ruff format` on `themes.py`.

(Item 1 I confirmed on `c9e050b`. Items 2 and 3 are carried from the previous
pass; I re-read the attribute names this round but did not run the page.)

---
*Verdict: Comment (PR is still a draft). Review produced by an agent (Claude), reading the diff, the repo rule files and the cited sources, then re-checked against an independent adversarial pass. Posted as a comment rather than a blocking review.*

Issue #447, opened 2026-08-30
[6 points each] Call for real Benchmark Radar use cases and report collaborators / 征集真实使用案例与报告协作者


ktwu01:
如果你懒得阅读下面的内容，请直接微信联系ktwu001，约10min电话

## Help turn Benchmark Radar into something people can use / 帮我们把 Benchmark Radar 变成真正可用的工具

The [published Benchmark Radar v0.9.0 technical report](https://doi.org/10.5281/zenodo.22167102) is frozen. We are improving the next technical-report draft now; accepted paper revisions will be merged into the paper repository, with reviewed versions linked back here. / [已发布的 Benchmark Radar v0.9.0 技术报告](https://doi.org/10.5281/zenodo.22167102)已经冻结。我们现在专注于改进下一版技术报告草稿；通过的论文修改会合并进论文仓库，再把审核后的版本更新到这里。

The report currently does **not** claim a recent benchmark trend. We need reproducible analysis and real use cases before making stronger claims. / 当前报告**没有**声称近期 benchmark 存在趋势。要得到更强的结论，我们需要可复现的分析和真实使用案例。

### Contribute to the paper / 参与论文协作

The manuscript now lives in **[benchmark-radar-paper](https://github.com/ktwu01/benchmark-radar-paper)**. Benchmark Radar includes a reviewed version as a Git submodule at `docs/technical-report/latex`. Issues, claims, and the points ledger stay in **this repository**. / 论文已迁到 **[benchmark-radar-paper](https://github.com/ktwu01/benchmark-radar-paper)**。Benchmark Radar 通过 Git submodule 引用已审核的论文版本，本地路径仍是 `docs/technical-report/latex`。任务、认领和积分账本继续放在**本仓库**。

- **Current draft / 当前草稿：** [Read `main.pdf`](https://github.com/ktwu01/benchmark-radar-paper/blob/main/main.pdf).
- **Edit the paper / 修改论文：** [`main.tex`](https://github.com/ktwu01/benchmark-radar-paper/blob/main/main.tex), [`references.bib`](https://github.com/ktwu01/benchmark-radar-paper/blob/main/references.bib), and [`figures/`](https://github.com/ktwu01/benchmark-radar-paper/tree/main/figures).
- **Writing and build guide / 写作与构建说明：** [paper README](https://github.com/ktwu01/benchmark-radar-paper#write-together-in-overleaf).
[...]

PR #639, opened 2026-09-13
Capability-theme coverage analysis: report-ready evidence (issue #447)

Jeffyaoliang:
Report-ready evidence for the collaboration call (#447): a reproducible
capability-theme coverage analysis of the agent-benchmark literature.

**Inputs are two public files** — the XBsleepy digest's published index and
this repo's committed snapshots — so every number below is checkable without
credentials or network.

## What the analysis shows

- The radar's snapshot taxonomy counts records by **type** (9,439 benchmark /
  6,744 evaluation / 6,240 dataset / 2,021 agentic sightings); the digest
  supplies the layer it does not carry: **which capability** an agent
  benchmark tests. Complementary, not overlapping.
- Corpus: **1,132 agent-benchmark papers** (announced 2026-01 .. 2026-09),
  1.94 capability labels per paper.
- Concentration: planning (38% of papers) + tool-use (36%) + safety (26%) =
  **52% of all labels**; strongest pairs planning+tool-use (171),
  safety+tool-use (124).
- Open areas: **embodied 38 papers (3%)**, workplace 20, mobile 14,
  recursive self-improvement 3 — the sparse end is where a new benchmark
  (or a new source) adds the most coverage.
- Trend: monthly tagged arrivals grew from 31 (Jan) to a May peak (91) and
  stayed elevated through August.

## Contents

- `analyze.py` — reproducible script (no network, no credentials)
- `coverage.md` — generated tables
- `themes_coverage.svg` — papers per capability theme
- `REPORT_SECTION_DRAFT.md` — draft technical-report section built on these
  numbers, ready to adapt

## Why here

The digest's capability layer enters the radar through #632 and is browsable
via #638; this analysis quantifies what that layer covers and where it is
thin. Per #447 — "Technical-report coauthorship additionally requires a
substantial intellectual contribution used in the report" — I'd like to
propose this as candidate evidence, and to adapt the draft section into the
report if the maintainers want it in a future version. Happy to reshape the
analysis (different cuts, more sources, other figures) to fit whatever the
report needs.

Related: #380, #632, #638


ktwu01 (changes requested):
Thank you for putting the analysis and draft together. I can’t recommend merging it as report-ready evidence yet.\n\nThe script reads mutable current snapshots and an unpinned external digest file rather than the frozen v0.11.0 report inputs and hashes. It also compares repeated snapshot category sightings with distinct digest papers, so the reported quantities do not share a unit. The draft points readers to , which is not on .\n\nPlease bind the analysis to the audited frozen inputs, export the complete identity and per-source coverage, deduplicate to a defined record unit (or explicitly relabel it as a sightings time series), and remove the unlanded route claim. The exact merge with current  also fails All checks passed! at  lines 20, 68, and 83. Thanks.

ktwu01 (commented):

The script reads mutable current snapshots and an unpinned external digest file rather than the frozen v0.11.0 report inputs and hashes. It also compares repeated snapshot category sightings with distinct digest papers, so the reported quantities do not share a unit. The draft points readers to `/themes/`, which is not on `main`.

Please bind the analysis to audited frozen inputs, export complete identity and per-source coverage, deduplicate to a defined record unit (or explicitly relabel it as a sightings time series), and remove the unlanded route claim. The exact merge with current `main` also fails `ruff check` at `analyze.py` lines 20, 68, and 83. Thank you.

ktwu01:
Thanks for the reproducible script and the draft. Three things before this can
be report evidence:

1. **Inputs are not frozen.** `analyze.py:23-28` defaults `DIGEST_INDEX` to
   `~/daily-agent-benchmarks/docs/data/index.json`, a path outside the repo
   with no pin or hash, and `SNAPSHOT_DIR` to the live `data/snapshots`, which
   grows daily. The corpus has already moved from 55 to 58 snapshots since the
   numbers in `coverage.md` were produced, so nobody can reproduce them today.
   Bind both sides to frozen v0.11.0 inputs and record the hashes.
2. **The conclusion outruns the measurement.** The comparison shows what one
   classifier's selected digest covers. Low observed coverage in that digest
   does not establish a field-wide research gap, and the draft currently reads
   as though it does. Either add the full-catalog accounting so the denominator
   is the corpus rather than the digest, or restate the finding as a statement
   about this digest's coverage.
3. **`REPORT_SECTION_DRAFT.md:20` routes readers to `/themes/`**, which is not
   on `main`; it lives in #638, still a draft. Drop the claim or gate it on
   that PR.

`ruff check` fails on three counts: `F401` unused `datetime` import at
`analyze.py:20`, and `E501` at `:68` (106 chars) and `:83` (105 chars).
`ruff check` is the first command in the CI sequence, so this stops the run
before anything else executes.

Withdrawing one thing I had queued: the counting-unit relabel. `coverage.md:33`
already heads the column "Sightings" and the draft says "category sightings",
so the distinction between repeat sightings and distinct papers is already
made. That was my misread.

---

PR #662, opened 2026-09-17
Break the saturation record-holder tie the way the frontier does

winklemad:
## The bug

`_best_row` (`src/benchmark_radar/benchmark_scores.py`) resolves a tied maximum by input order, so a benchmark's `saturation.best_*` record-holder can disagree with its own `historical_best_frontier` — and can flip when the source rows are reordered.

`_best_row` sorted only by `(value, reported_at)`. A stable sort leaves an exact tie to YAML row order — the very failure `_historical_best_frontier` was written to avoid ("*A stable observation id breaks an exact same-date/value tie without depending on YAML row order*") and that `test_a_gain_endpoint_does_not_depend_on_a_model_name` (Codex P2) pins for the gain endpoints. `_best_row` had no such disambiguator, so it reintroduced the order-dependence for `saturation.best_*`. Two divergences from the frontier:

1. **Same-date value tie** → resolved by row order instead of the frontier's smallest `observation_id`.
2. **Higher-is-better date direction** → `_best_row` picked the *latest* date the max was reached; the frontier records the *earliest* (a later equal value "is not a new record").

## It's live in the shipped data

`build_score_progression(...)["benchmarks"]["frontier_challenge"]`:
- `saturation.best_model` = **`GPT-5.6 Sol`**
- `historical_best_frontier.points[-1].model` = **`Grok 4.6`**

Both describe the identical record (value `20.6`, `2026-08-25`), yet the saturation block and the record-progression chart credit different models. `saturation.best_*` is rendered into the `closing_headroom` insight sentence (`insights.py`) and exported to `radar.json` (`snapshots.py`), so the site shows the contradiction directly. The best *value* and *headroom* are correct — only the record-holder attribution was wrong, so no score/aggregate is numerically off.

## Fix

Break the tie exactly as `_historical_best_frontier` does: best value (directional), then the **earliest date** it was reached, then the **smallest `observation_id`**. `saturation.best_*` now agrees with the frontier's final record and no longer depends on row order. The same `_best_row` feeds the gain endpoints, so their `from_model`/`to_model` become order-independent too.

After the change, `frontier_challenge` `saturation.best_model` is `Grok 4.6` (the frontier's holder), and **no shipped benchmark disagrees**.

## Tests / validation

Added `test_saturation_best_holder_matches_the_frontier_on_a_tie` — two models tied at the record value must name the same holder as the frontier and not depend on row order (fails before, passes after). `ruff check` / `ruff format --check` clean; `pytest tests/test_benchmark_scores.py` green (34). The `saturation_audit._best_row` variant is a separate function with its own golden-pinned tie key and is untouched.


winklemad:
`330226` Claude

Per AGENTS.md: this PR was authored with a coding agent (Claude). I've read and understand the diff, verified it against the shipped `frontier_challenge` data and with the added regression test, and I'll revise it in review like any other contributor.

ktwu01:
Verified the fix matches the frontier's tie-break convention and that every `_best_row` input path carries `observation_id` (assigned at load, preserved on points), so the new key can't `KeyError` on shipped data. The regression test pins exactly the reported property (order-independence plus frontier agreement). Approving; please just confirm the test suite is green on a clean checkout since no CI checks are visible on this PR.

winklemad:
Thanks for the quick review and merge! Confirmed `pytest -q` is green on a clean checkout locally — the new order-independence test plus the full suite. Appreciate you verifying the `observation_id` paths.

PR #664, opened 2026-09-17
Keep assign_slugs unique against natural slugs, not just per-base

winklemad:
## The bug

`assign_slugs` (`src/benchmark_radar/catalog.py`) promises a unique slug per key ("*deterministic suffixes when two collide*"), but it only tracks a per-base counter and blindly appends `-{count}`. It never checks that the disambiguated slug isn't already another key's **natural** slug, so a base that needs disambiguating to `X-2` collides with a distinct key whose natural slug is already `X-2`.

```python
count = used.get(base, 0) + 1
used[base] = count
assigned[key] = base if count == 1 else f"{base}-{count}"
```

Minimal repro: `assign_slugs(["vending bench", "vending-bench-2", "vending:bench"])` emits `vending-bench-2` twice.

## Why it matters

The slug is a record's shard filename and its catalog URL, and it must be unique:
- `catalog_shards.write_shards` writes by slug → a duplicate silently **overwrites one record's detail shard** (a benchmark's page is lost / serves the wrong record). AGENTS.md: "*Missing measurements must not remove corpus records.*"
- `query._load_index` treats duplicate slugs as a hard `invalid_data` error ("*benchmark index keys and slugs must be unique*"), so a single collision takes the whole catalog query surface (CLI + HTTP `show`/`search`/`status`) down.

**Honest scope:** this is **latent** — the current crawl doesn't trigger it (`test_slugs_are_filename_safe_and_unique` confirms uniqueness holds today). But it's reachable as the corpus grows: the trigger is a benchmark id whose natural slug ends in `-N` coexisting with punctuation/space variants of the same base — realistic for an aggregator that already carries versioned ids like `vending-bench-2`. When it hits, the failure is severe (lost record + total query outage), so this hardens the invariant the function already promises before it bites.

## Fix

Track the emitted final slugs in a set and bump the suffix until the candidate is globally free, instead of trusting `f"{base}-{count}"`. Still deterministic (processed in sorted key order) and unchanged for the common case (two variants of one base still get `base` / `base-2`).

## Tests / validation

Added `test_assign_slugs_disambiguates_without_colliding_with_a_natural_slug` (fails before, passes after). `ruff check` / `ruff format --check` clean; `pytest tests/test_catalog.py` green (57).


winklemad:
`330226` Claude

Per AGENTS.md: authored with a coding agent (Claude). I've read and understand the diff, reproduced the collision and the fix with the added regression test, and I'll revise it in review like any other contributor.

ktwu01:
Clean fix, verified the repro logic and that the common case (`base`, `base-2`) is unchanged, so no existing shard filenames churn. Approving. Two non-blocking notes: consider asserting the exact slug mapping in the new test (natural `vending-bench-2` keeps its slug), and note that `src/benchmark_radar/catalog_opencompass.py:334-347` carries the same per-base-counter pattern for adapter-local slugs if anyone wants a follow-up. Please confirm CI is green since no checks are visible here.

winklemad:
Thanks! Confirmed green locally on a clean checkout (`ruff check`/`ruff format --check`/`pytest -q`). Good calls on both notes — agreed the test reads better asserting the exact slug mapping (`vending-bench-2` keeps its natural slug), and I spotted the same per-base-counter pattern in `catalog_opencompass.py:334-347`; happy to send a small follow-up for that one.

PR #667, opened 2026-09-19
feat(ui): improve research navigation and unify site styling

lynnzc:
## Summary

Improve the existing site's navigation, typography, spacing, and controls to make benchmark research easier to browse.

- Apply shared styling to the dashboard, articles, and benchmark pages using the original header and radar logo.
- Add a research field browser at `/research/`, with two-level filters, source evidence, and export, through the normal site build.
- Carry field selection between Frontier and score history while keeping benchmark search open to the full catalog.
- Preserve existing routes and use the canonical index with on-demand detail loading.

## Validation

- Full CI sequence passed from a clean checkout with submodules initialized: lint, formatting, catalog normalization, classification, data release, and **1,386 tests**.
- **18 interaction checks** passed for navigation, field filters, search, evidence details, and language switching.


lynnzc:
330226 — GPT

This PR was prepared by a GPT-series coding agent on behalf of @lynnzc, following AGENTS.md and CONTRIBUTING.md. The complete clean-checkout CI sequence passed locally (1,385 tests), along with 22 UI checks. The uploaded head has the same Git tree as the tested checkout. Desktop and mobile visual review remains outstanding; the proposal is opt-in and does not change the production deployment.

ktwu01:
Checked corpus compliance: the default field `"all"` returns the full index
(`site/assets/fields.js`, `matchesField`), untagged records land in `"other"`
rather than being dropped, and Saturation search bypasses both the cutoff and
the field filter, which matches `principle.md`. The gitignored `/research/`
page and the on-demand shard loading look right, and no `site/data/*.json` was
hand-edited.

Three asks:

1. `src/benchmark_radar/site_pages.py:745` and `:749` read
   `Path("site/index.html")` and `Path("site/assets/app.js")` CWD-relative
   while the same function writes to `staging` / `output_dir` (`:735`, `:771`).
   These reads are unconditional, so a caller invoking this from any other
   working directory gets a `FileNotFoundError`, not a degraded page. Resolve
   them against the repo root, or document the constraint at the signature.
2. `research.js:111` renders `r.metric` and `r.direction` in the Record
   metadata panel, but `research_page.py` never emits either field, so every
   record shows "Metric not recorded / Direction unknown". The fallbacks keep
   it from breaking, which is also what makes it easy to miss. Either populate
   both from the shard or drop the panel.
3. Please complete the outstanding desktop and mobile visual review and confirm
   the full six-step run on a clean checkout.

---
*Verdict: Comment. Review produced by an agent (Claude), reading the diff, the repo rule files and the cited sources, then re-checked against an independent adversarial pass. Posted as a comment rather than a blocking review.*

lynnzc:
Addressed the two code findings: generated benchmark pages now resolve shared chrome from the repository path, with a regression test for a non-root working directory; the Research details dialog no longer renders unsupported record metadata. The six-step CI sequence passed in a clean detached worktree with submodules initialized (1,387 tests). The updated preview is https://benchmark-radar-study.lynnzc.chatgpt.site. Desktop/mobile visual confirmation remains pending because the preview browser could not start in this environment.

PR #674, opened 2026-09-20
fix(site): preserve search query in CLI handoff

### LINKED ISSUE #487: [2 Points] tricky redirect workflow
### What happened / 发生了什么

when usr search for RSI
they will be at
https://benchmark-radar.org/?date=all&q=RSI
and they will see
```
This search covers all dates. Still want today's results? [Search today](https://benchmark-radar.org/?date=2026-08-31&q=RSI).

More than 10 results. [Use the CLI version to export all data.](https://benchmark-radar.org/#cli)
```

and then, when they click  [Use the CLI version to export all data.](https://benchmark-radar.org/#cli)
everything about the search keyword DISAPPEARED AND
it only shows only https://benchmark-radar.org/#cli

so basically we wanna a dynamic page about CLI
that goes dynamically w/ usr query

e.g., it will show https://benchmark-radar.org/#cli&q=RSI

and the prompt of agent will be like
```
Set up Benchmark Radar for local benchmark search. Follow
https://github.com/ktwu01/benchmark-radar/blob/main/skills/benchmark-radar/SKILL.md
to install the CLI and consumer Skill, initialize the local data, and verify the setup. 
```

AND add to the prompt "then Search for RSI use this CLI and skill"

### Where did you see it / 在哪里看到的

_No response_

bigdu332:
## Summary

- Preserve the Today search query when opening the CLI handoff.
- Show a copyable agent prompt that continues the same search.
- Keep direct `/cli/` links working without a query.

Closes #487

## Design note

Commit `48413f8` removed the generic setup prompt because the consumer Skill owns setup end to end. This PR keeps that boundary: only the query-specific `/cli/?q=...` handoff includes an agent prompt. It points to `SKILL.md` for setup and adds the current search request, rather than duplicating installation steps on the page. The bare `/cli/` route remains install-only.

## Testing

- `ruff check .`
- `ruff format --check .`
- `benchmark-radar normalize-catalog`
- `benchmark-radar classify`
- `benchmark-radar build-data-release`
- `pytest -q` — 1384 passed


ktwu01:
Thanks, this tracks issue #487 closely. The generated prompt matches the
wording in the issue, including the setup sentence and the "then search for X"
follow-on, and `/cli/?q=` is the current path-style equivalent of the
`#cli&q=` the issue sketched. Round-tripping, the `state.view === "today"`
gate, and the untouched bare `/cli/` route all check out, and the helpers it
leans on are safe for this use: `t()` takes a params object (`app.js:237`),
`element()` drops falsy children so `query && element(...)` renders nothing
when there is no query, and `.copy-text` is already `white-space: pre-wrap`
so the multi-line prompt keeps its breaks.

One thing to settle before merge, and it is a maintainer call rather than a
defect in this diff:

**This re-lands an affordance that was deliberately removed, and the
chronology does not resolve which decision is current.** Commit `48413f8`
("reduce local CLI setup to the install command", 2026-09-02) deleted the
earlier `CLI_AGENT_PROMPT` with the rationale "Setup logic does not belong in
copy the reader has to paste on faith. The Skill now owns setup end to end."
Issue #487 asks for a query-specific prompt, but it was opened 2026-09-01, the
day *before* that commit, so it does not supersede the removal on its own.

The two may well be reconcilable: #487's prompt delegates the actual setup to
the `SKILL.md` URL rather than spelling the steps out, which is close to what
`48413f8` wanted. If that is the reading, please say so in the PR body so the
next reader sees a decision reconciled rather than silently undone. The diff
currently also deletes the comment that recorded `48413f8`'s reasoning, which
is the line that would have explained this.

Two smaller points:

- If an agent opened this PR, `AGENTS.md` asks for a comment carrying the
  marker `330226` and the model series. I am asking rather than assuming; a
  branch name is not evidence either way.
- No length bound on `q`, so a very long query produces a very long prompt in
  the card. `JSON.stringify` keeps it on one escaped line, so this is cosmetic.

---
*Verdict: Comment. Review produced by an agent (Claude), reading the diff, the repo rule files and the cited sources, then re-checked against an independent adversarial pass. Posted as a comment rather than a blocking review.*

bigdu332:
Thanks for the careful review. I updated the PR body and the nearby code comment in `07d6c7d` to record the design rationale.

My reading of `48413f8` is that setup steps belong in the consumer Skill rather than in page copy. This PR keeps that boundary: the bare `/cli/` route remains install-only, while the query-specific handoff points the agent to `SKILL.md` for setup and passes along the current search request. I understand this is ultimately a maintainer product decision; if the earlier removal was intended to rule out even query-specific agent prompts, please let me know and I can adjust.

For provenance: I clicked Create pull request manually after Codex generated the original implementation. Codex also prepared and pushed the follow-up commit and helped write this reply. The code contribution is substantially agent-authored, so I am including the repository annotation for transparency: **330226 GPT**.

